PR #9, opened 2024-03-25
Optimizing the format of Curl command on first prompt

aertoria:
Optimizing the format of Curl command on first prompt

The current formatting of curl command is confusing. Making reader think it's missing enclosing


MarkDaoust (approved):

aertoria:
thanks!

PR #17, opened 2024-03-29
Swap File API Cookbooks to use SDKs

TYMichaelChen:
Currently just a draft.

DO_NOT_SUBMIT


MarkDaoust:
Trying this out.

For the Video example .. 1/5 uploads are failing:

https://screenshot.googleplex.com/5G2UKtgH6TA8y84

Quick hack retry would be:

```
  for n in range(5):
    try:
      response = genai.upload_file(path=file.file_path)
    except Exception as e:
      error = e
      print(e, "retrying")
      continue
    break
  else:
    raise error
    ```
    
   Wait, now it's down to 1/100.
   
   ??

**Comment by TYMichaelChen** [comment]
Yup it’s a known issue for just my project, if you switch api keys to others this should not be an issue

**Comment by MarkDaoust** [comment]
Faster uploads I tried this in colab, but the print-outs were basically sequential, no faster:

```
import concurrent.futures

def do_upload(file):
  print(f'Uploading: {file.file_path}...')
  for n in range(5):
    try:
      response = genai.upload_file(path=file.file_path)
    except Exception as e:
      error = e
      print(e, "retrying")
      continue
    break
  else:
    raise error

  file.set_file_response(response)
  return file

with concurrent.futures.ThreadPoolExecutor(max_workers=10) as executor:
    result = list(executor.map(do_upload, files_to_upload))
 ```

MarkDaoust (approved):

TYMichaelChen:
Note you need to install the SDK via the github url: `!pip install -U git+https://github.com/google/generative-ai-python`

PR #40, opened 2024-04-09
Add top-level .gitignore, delete .DS_Store

williamito:


markmcd (approved):
Nice, thanks! I added some more ignorables too.

PR #44, opened 2024-04-10
Force bash for sample.sh

markmcd:
The script uses bash-specific semantics (e.g. `[[`)  but does not force bash.

This causes silent partial-failures on terminals where /bin/sh is not bash.

### DIFF
--- a/quickstarts/file-api/sample.sh
+++ b/quickstarts/file-api/sample.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
 # Upload a file using the GenAI File API via curl.
 api_key=""


TYMichaelChen (approved):
Ah thank you!

markmcd:
Thanks for the quick review! This was quite the debugging session... For posterity:

`dash` doesn't have `[[`, so branch just gets skipped:
```sh
  if [[ ${i} -eq ${NUM_CHUNKS} ]] ; then
    # For the final chunk, specify "finalize".
    upload_command="${upload_command}, finalize"
  fi
```

So the upload is never finalised, and the File `GET` call doesn't show the upload status, the only symptom is a lack of `displayName`, we otherwise see a normal response.

PR #46, opened 2024-04-10
Make sample.sh executable

markmcd:
Git preserves the executable bit, so turn it on for this helpful script.


TYMichaelChen (approved):
thanks 🙏

Issue #58, opened 2024-04-12
curl URLs should be quoted. 


sahinf:
Problem:
I attempted to run a curl [command](https://github.com/google-gemini/cookbook/blame/9eb52260f979aa8339e6d1eb77323faa178bbd78/quickstarts/rest/Prompting_REST.ipynb#L140C14-L140C19) and got a "no matches found" error

Explanation:
This is because the shell that I ran this command from, zsh, explicitly raises an error when it cannot expand a wildcard. 
In this case, it is attempting to expand the `?` used in the query string `?key=$GOOGLE_API_KEY`

AFAIK this only affects zsh users because other shells like bash will [pass the unexpanded wildcard](https://www.gnu.org/software/bash/manual/html_node/Filename-Expansion.html) pattern instead of raising an error like zsh. 
Example:
![image](https://github.com/google-gemini/cookbook/assets/6597168/d4fc5eab-8f01-420a-9ce8-40476496c6b2)

PR #59, opened 2024-04-12
Quote curl URLs

sahinf:
Fixes issue #58 

To find all instances of curl that needed to be fixed, I  grepped regex `"!?curl\s+(-o\s+\S+\s+)?(https?://\S+)` and then added quotes to the 24 instances of unquoted curl URLs.

Please let me know if you think I missed a spot or if you think this is completely unnecessary. The referenced issue isn't really a bug, but rather an inconvenience for lazy zsh users like me.

### DIFF
--- a/examples/Guess_the_shape.ipynb
+++ b/examples/Guess_the_shape.ipynb
@@ -78,9 +78,9 @@
       },
       "outputs": [],
       "source": [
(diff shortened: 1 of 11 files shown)


markmcd (approved):
LGTM.

PR #85, opened 2024-04-19
Add `nblint` and `nbfmt` checks to pull requests

markmcd:
These args should work for the cookbook, though there are some KIs:

* The button-check looks for a table with `class="tfo-notebook-buttons"`, which we have stripped from cookbook recipes.
* Lots of notebooks use "we" in an acceptable way - e.g. as LLM output, but we don't have a way to suppress this check per-cell.
* License & copyright checks are disabled. We don't have copyright and licenses headers added to content yet, though we are required to. We should add them before enabling the checks.

We should also set this up so that `nbfmt` runs and commits automatically too but for now I'm just copying our existing setup.


markmcd:
Added the license/copyright checks back in after #91. Attempting a fix for the first issue in https://github.com/tensorflow/docs/pull/2304

MarkDaoust (approved):

PR #89, opened 2024-04-21
Create Internet_Access_Function_Calling.py

sankeerthrao:
Adding internet access to Gemini via the Function Calling API.

### DIFF
--- a/examples/Internet_Access_Function_Calling.py
+++ b/examples/Internet_Access_Function_Calling.py
@@ -1,0 +1,96 @@
+import os
+from random import randint
+from typing import Iterable
+import google.generativeai as genai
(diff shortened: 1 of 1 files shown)


nafSadh:
Thanks for sharing this cool demo from AGI house.

MyNameIsTarric1:
Hi, if i use your code the model always answer saying: "I apologize, but I'm unable to access real-time information at the moment."
Why does it not searching on internet?

markmcd:
Are you able to convert this into a IPython notebook? You can use the template we have [here](https://github.com/google-gemini/cookbook/blob/main/templates/Template.ipynb).

PR #91, opened 2024-04-23
Add license headers and copyright to notebooks

markmcd:
Ideally these would be collapsed by default. LMK if you have any suggestions - Colab didn't seem to export anything and I didn't look much farther than that.


random-forests:
Thank you!

random-forests (approved):

PR #95, opened 2024-04-26
Update Authentication_with_OAuth.ipynb

KaylaMLe:
The equal sign after --scopes and the single quotes before the urls were being interpreted as part of the first scope URL. The last single quote was being interpreted as part of the third scope URL.

This caused a 400 error with the message "Some scopes invalid" and prevented sign in. Running the command without the = or ' was successful.


KaylaMLe:
Also the backslashes as line breaks in that same command tripped me up for a moment. I thought the Google Cloud Shell would automatically interpret the whole block as a single command if I just copied and pasted it. Though, I'm not familiar enough with notebooks to know how the command could be formatted better.

markmcd (approved):
Thanks for the patch! I added some unrelated changes to fix the lint error.

PR #98, opened 2024-05-01
switch from openai to gemini

satvik314:
Added notebook guide to use Gemini API with full OpenAI compatibility using Portkey.


markmcd:
Thanks for the contribution.  Before we do a deep review, we have a few questions about the content. It might be easier to create a new issue where we can work out what guides to include and what they look like.

My initial questions are:
* Is it possible to run the Portkey system locally instead of using a remote service? This way users can try it out without giving their API key to a third party.
  * If not, would you mind if we put a warning on the part where they share their API key?
* Can you provide some stats on the platform's popularity? See our [contrib guide](https://github.com/google-gemini/cookbook/blob/main/examples/CONTRIBUTING.md), but we are only accepting contributions for widely adopted platforms, with some exceptions for technical novelty. Public numbers are fine.
* Can you ensure the guide is focused on a single topic (e.g. "using a compatible interface" is one topic, "observability with portkey" is another)
  * Focused content is more useful to users, and it will help us review each guide on its own merits.

We also have [a template](https://github.com/google-gemini/cookbook/blob/main/templates/Template.ipynb) you can use now, it includes some authoring guidelines. And you'll need to rebase your change to HEAD so you pick up the latest GH actions.

PR #126, opened 2024-05-11
Update "space" issue in the header in all .ipynb files

doggy8088:
There is a "space" issue in the header.


markmcd (approved):
Thanks. For context, this was the standard template we used until very recently: https://github.com/tensorflow/docs/pull/2307

PR #128, opened 2024-05-12
The "pip install -U -q google.generativeai" should change to "pip install -U -q google-generativeai" for consistency

doggy8088:
There are some documents write this way:

```sh
pip install -U -q google.generativeai
```

and some others are write this way:

```sh
pip install -U -q google-generativeai
```

I just references from here: https://ai.google.dev/api/python/google/generativeai

The package name should be consistent. So I updated it to `google-generativeai`.

### DIFF
--- a/examples/Anomaly_detection_with_embeddings.ipynb
+++ b/examples/Anomaly_detection_with_embeddings.ipynb
@@ -83,7 +83,7 @@
         }
       ],
       "source": [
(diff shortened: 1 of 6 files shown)


markmcd (approved):
Definitely :) Thank you for this!

Looks like the nblint issue is pre-existing. Would be nice to fix but not relevant to this change, so I'll override it.

PR #150, opened 2024-05-17
add basic reasoning prompting example

shilpakancharla:


examples/prompting/Basic_Reasoning.ipynb
```diff
@@ -0,0 +1,217 @@
+{
```
markmcd, line 1: <p>This link looks wrong - it has /quickstarts in it.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google-gemini/cookbook/pull/150/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

markmcd, line 1: <p>Gemini is the consumer chat app. Make sure you only talk about "Gemini" when you mean the app (this applies everywhere in every guide).</p><p> </p><p>For Gemini API examples, you can say "The Gemini API" or just "the models" is often enough.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google-gemini/cookbook/pull/150/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

markmcd, line 1: <p>Can we provide any narrative text to go with this?</p><p> </p><p>Not sure if we need to or not - honestly this is pretty clear as-is - but even a short explanation for why the template &amp; problem are separated (e.g. "First you set the template, this is reusable across different problems...")</p><p> </p><p>And this part should go in the system instruction, with the problem passed directly to generate_content.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google-gemini/cookbook/pull/150/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

markmcd, line 1: <pre class="ql-syntax" spellcheck="false">Line #1.    math_problem = template.format(problem="Given a triangle with base 𝑏=6 and height h=8, calculate its area")
</pre><p> </p><p>Nit: Suggest using either regular ascii or unicode characters for math symbols, don't mix them.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google-gemini/cookbook/pull/150/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

shilpakancharla: With the recent push I think the mix of characters has been updated -- but let's double check once more how it appears.

examples/prompting/Basic_Reasoning.ipynb
```diff
@@ -0,0 +1,239 @@
...
```
markmcd, line 1: <pre class="ql-syntax" spellcheck="false">Line #11.    Problem: """
</pre><p> </p><p>Putting a trailing clause at the end of the system instruction assumes that the system instruction is concatenated with the input contents. I don't know that we can or should assume that.</p><p> </p><p>If this works fine without this line, then remove it. If it's needed, move it to the input text.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google-gemini/cookbook/pull/150/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

markmcd (dismissed):
LG other than that one little suggestion.

examples/prompting/Basic_Reasoning.ipynb
shilpakancharla: Just removed and it seems to be working fine. Thanks Mark!

MarkDaoust (dismissed):
LGTM if we can change the wolf-goat-cabbage problem.

PR #165, opened 2024-05-21
Add role prompting example

shilpakancharla:


MarkDaoust (changes requested):
Please use system_instruction.

MarkDaoust (approved):

Abhishek0075:
> Please use system_instruction.

When i used system_instruction it is giving the TypeError: __init__() got an unexpected keyword argument 'system_instruction'
What should i do ?

Can you show if there is any problem ?

Code : 

```

import os
from dotenv import load_dotenv

load_dotenv()
genai.configure(api_key=os.environ["GEMINI_API_KEY"])

model = genai.GenerativeModel(
    "models/gemini-1.5-flash",
    system_instruction="You are a cat. Your name is Neko.",
)
prompt = "Write a story about a magic backpack."

response = model.generate_content(prompt)

print(response.text)
```

PR #168, opened 2024-05-21
Add prompting example for self-ask

shilpakancharla:


MarkDaoust (changes requested):
Can we link to the wikipedia search examplke at the end?: 

https://github.com/google-gemini/cookbook/blob/22ba52659005defc53ce2d6717fb9fedf1d661f1/examples/Search_reranking_using_embeddings.ipynb

PR #177, opened 2024-05-28
Add text summarization JSON example

shilpakancharla:
Uses Gemini 1.5 Flash model with `response_schema`. Please let me know if this should use Gemini 1.5 Pro.


MarkDaoust (approved):
There's no way to set tyhe fields to required right now. I shopuld fix that.

PR #178, opened 2024-05-28
Add entity extraction JSON example

shilpakancharla:
Using Gemini 1.5 Flash with `response_schema`.


MarkDaoust (approved):
No example with "response_schema"?

PR #180, opened 2024-05-29
Point gemini-1.5-flash-latest as default model for all quickstart notebooks

lucianommartins:
Using gemini-1.5-flash-latest as default model for quickstart notebooks.


quickstarts/contrib.md
```diff
@@ -0,0 +1,45 @@
+# Contributing to the Gemini API Cookbook
```
MarkDaoust, line 1: Isn't this sort of information usually put at the top level in a `CONTRIBUTING.md` file?

MarkDaoust (changes requested):
LGTM, but can we put the contribution instructions is it's own file?

quickstarts/contrib.md
MarkDaoust: Oh.. it exists.

https://github.com/google-gemini/cookbook/blob/main/CONTRIBUTING.md

MarkDaoust (approved):

PR #181, opened 2024-05-29
Point gemini-1.5-flash-latest as default model for all notebooks on examples folder

lucianommartins:
Using gemini-1.5-flash-latest as default model for all notebooks under examples folder.


MarkDaoust (dismissed):
Examples are less important than the quickstarts, so LGTM.

MarkDaoust (approved):

PR #187, opened 2024-05-30
Add a notebook test script.

MarkDaoust:


MarkDaoust:
It's still pretty rough, but it gets the notebooks run.

Some notebooks only run with api_key, others only with oauth. We'll need to fix this.

We should probably look at using colab's docker: 

https://twitter.com/GoogleColab/status/1663594080436375558

Then run in GH actions or kokoro.

https://docs.github.com/en/actions/creating-actions/creating-a-docker-container-action

PR #192, opened 2024-06-03
Use genai.protos instead of google.ai.generativelanguage

MarkDaoust:
Use `genai.protos` to access the raw protos.


markmcd (approved):
Looks like a couple of search/replace errors got in, so they're fixed. LMK if I changed something I shouldn't have.

PR #197, opened 2024-06-07
Point to the wrapper, not the proto.

MarkDaoust:


markmcd (approved):

MarkDaoust:
Wait. This class is not exposed in the docs. Pls fix.

PR #204, opened 2024-06-10
Remove references to pro-vision

MarkDaoust:


quickstarts/rest/Prompting_REST.ipynb
```diff
@@ -379,7 +379,7 @@
       "source": [
         "%%bash\n",
         "\n",
-        "curl \"https://generativelanguage.googleapis.com/v1beta/models/gemini-pro-vision:generateContent?key=${GOOGLE_API_KEY}\" \\\n",
+        "curl \"https://generativelanguage.googleapis.com/v1beta/models/gemini-1.5-flash:generateContent?key=${GOOGLE_API_KEY}\" \\\n",
```
markmcd, line 382: Can you re-run these cells and save the outputs? I'm not sure how strict we need to be on this but since it seems simple it'd be good to err on the side of accurate, reproducible docs.

markmcd (approved):
LG but in case it's important, the outputs should be reproducible.

markmcd (dismissed):

markmcd (approved):

markmcd (approved):

PR #213, opened 2024-06-18
Add context caching quickstart

markmcd:
I used the content from the Apollo 11 example as it is quick to upload & start using (vs a video which needs a wait loop).

This covers cache setup, management & lifecycle, plus demos how to use both chat and generate_content, and shows some token accounting.


random-forests:
Thank you! This is great. I made minor changes to the prompts (I didn't tell the model it was analyzing a nasa transcript in the SI to make the task more challenging).

PR #227, opened 2024-06-28
Update Code_Execution.ipynb

pablosanchez84:
Improved prompt sample


markmcd:
Thanks for the PR - could you make sure the cells are re-run so that the output reflects the updated input?

markmcd:
I think this change is superseded by #230. Please shout out if you disagree, or just create a new PR.

PR #238, opened 2024-07-11
Adding a "What's new?" section in the main README.

Giom-V:
We can discuss where exactly to add it in the README. On one hand I think it would make sense to be at the beginning, on the other hand that is not useful for newcomers.


random-forests (approved):

LIPIKASTUDIO354612 (commented):
Photo editor and Google news browsing and Analytics

PR #245, opened 2024-07-22
Call Gemini in OpenAI format (via LiteLLM) - tool calling, image url's, etc. in docs

krrishdholakia:
Adds docs on calling gemini in the openai format

- via litellm sdk 
- via litellm proxy (allows users to set budgets + track usage across llm providers)

### DIFF
--- a/examples/litellm/proxy.md
+++ b/examples/litellm/proxy.md
@@ -1,0 +1,176 @@
+# Use Gemini in OpenAI Format (via LiteLLM Proxy)
+
+Use LiteLLM Proxy to track usage + set budgets call Gemini/Anthropic/OpenAI etc.
+
+
+Works for [Google AI Studio](https://docs.litellm.ai/docs/providers/gemini) + [Vertex AI](https://docs.litellm.ai/docs/providers/vertex) calls.
+
+## Sample Usage
+
+### Pre-requisites
+* `pip install -q google-generativeai`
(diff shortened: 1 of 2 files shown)


krrishdholakia:
cc: @shilpakancharla @Giom-V

let me know if there's anything missing for merge!

markmcd:
Hey @krrishdholakia - thanks for the contribution. I think the idea is solid, and we'd be happy to have something showing how to use litellm in this cookbook, though I have some notes.

Given this is in Python, can you format it as a Python notebook? And can you focus the content specifically on how to use LiteLLM with the Gemini API?

If you want to put it all in one notebook you could, or separate notebooks per topic would be fine if they are detailed enough to justify it. A topic like "Getting started with LiteLLM and the Gemini API" would be great, or something showing the features of litellm would be even better - the idea of `track usage + set budgets` sounds compelling, if you could show the budgets and usage in the content it'd be excellent.

krrishdholakia:
got it @markmcd i'll update the model strings to correctly reference the gemini api

krrishdholakia:
Hi @markmcd updated with detailed examples for tracking usage + setting budgets. Also did some additional cleanup to reference gemini in the model strings. 

Is this good for merge?

markmcd:
> can you format it as a Python notebook? 

Can you address this?

PR #248, opened 2024-07-28
Update Authentication_with_OAuth.ipynb

robinroy03:
The other URL is deprecated and gives a 404 now.


markmcd:
I'm not sure this is the right fix. Models still must be named `tunedModels/something`, so the original syntax is still correct. If it is changed, the field should default to something containing the prefix `tunedModels/` so it's clear what belongs there.

PR #263, opened 2024-08-22
Add a template notebook.

markmcd:
This is for new contributors or Googlers to use as a base for new quickstarts or examples.

It's largely forked from the [TensorFlow template](https://github.com/tensorflow/docs/blob/master/tools/templates/notebook.ipynb), but I've updated the guidance to apply to the Gemini API and this cookbook.


markmcd:
Looks like the `nblint` finding is where I have explicitly called out using "we".

MarkDaoust (approved):

MarkDaoust:
Notebook lint is a false positive.

PR #271, opened 2024-09-04
Add an example using Enums

MarkDaoust:


markmcd (approved):
LGTM though I strongly disagree with the final cell output. Linzer > Raisin Oatmeal.

PR #276, opened 2024-09-10
Pin a minimum version for each notebook (0.7.2 default)

MarkDaoust:


random-forests:
Thanks for this. Hmm, question -- 

In colab, the version currently installed by default is 0.7.2.

```
import google.generativeai
google.generativeai.__version__

0.7.2
```

But running

`!pip install -qU \"google-generativeai>=0.7.2`

still causes colab to uninstall and reinstall the package. Is that just me? (Is there a way we can get a minimum version pinned for some notebooks that doesn't cause that, just so they run faster?) Thanks!

MarkDaoust:
I'm updating Colab to use 0.8.0 in cl/673024762.

random-forests:
Nice!

random-forests (approved):

Alexisloyolamend3z:
Hola amigos

PR #277, opened 2024-09-11
Close stale issues/PRs after 13 days

markmcd:
As discussed, reducing the close date to 13 days so we can review fortnightly. Also updated the scheduled time so it runs before our current meeting schedule.

I left the initial reminder to 14 days as I think the closure is the key event, but lmk if you think we should do both.


Alexisloyolamend3z:
Hola amigos

MarkDaoust (approved):

MarkDaoust:
Thanks!

Issue #66, opened 2024-04-15
File upload example for javascript returns No file found in request.


LindaLawton:
I tried to run the [file upload example for javascript](https://github.com/google-gemini/cookbook/blob/cf2dc7a3a851c5f553c2bf9ff884770719ca368d/quickstarts/file-api/sample.js#L25)

The file exits  as i added a check for it

    // Check if the file exists
    if (!fs.existsSync(filePath)) {
        console.log(`File '${filePath}' does not exist` );
    }

>GaxiosError: No file found in request.
    at Gaxios._request (C:\Development\Gemini\nodejs\gemini_samples_node\node_modules\gaxios\build\src\gaxios.js:144:23)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async JWT.requestAsync (C:\Development\Gemini\nodejs\gemini_samples_node\node_modules\google-auth-library\build\src\auth\oauth2client.js:405:18)
    at async run (C:\Development\Gemini\nodejs\gemini_samples_node\test2.js:30:36) {
  config: {
    url: 'https://generativelanguage.googleapis.com/upload/v1beta/files?uploadType=multipart',
    method: 'POST',
    userAgentDirectives: [
      {
        product: 'google-api-nodejs-client',
        version: '7.1.0',
        comment: 'gzip'
      }
    ],
    paramsSerializer: [Function (anonymous)],
    data: PassThrough {
      _readableState: ReadableState {
        objectMode: false,
        highWaterMark: 16384,
        buffer: BufferList { head: null, tail: null, length: 0 },
        length: 0,
        pipes: [],
        flowing: false,
        ended: true,
        endEmitted: true,
        reading: false,
        constructed: true,
        sync: false,
        needReadable: false,
        emittedReadable: false,
        readableListening: false,
        resumeScheduled: false,
        errorEmitted: false,
        emitClose: true,
        autoDestroy: true,
        destroyed: true,
        errored: null,
        closed: true,
        closeEmitted: true,
        defaultEncoding: 'utf8',
        awaitDrainWriters: null,
        multiAwaitDrain: false,
        readingMore: false,
        dataEmitted: true,
        decoder: null,
        encoding: null,
        [Symbol(kPaused)]: true
      },
      _events: [Object: null prototype] {
        prefinish: [Function: prefinish],
        error: [Function (anonymous)]
      },
      _eventsCount: 2,
      _maxListeners: undefined,
      _writableState: WritableState {
        objectMode: false,
[...]

PR #279, opened 2024-09-12
Use `path.join` for File API JS sample

markmcd:
Trying to help with #66, the issue is on Windows so this *might* help. It shouldn't hurt anyway.


Giom-V:
@markmcd Is that still relevant?

markmcd:
Yeah I guess it is. It should be replaced with the new SDK but until then this PR makes it work with Windows.

Giom-V (approved):

Giom-V:
> Yeah I guess it is. It should be replaced with the new SDK but until then this PR makes it work with Windows.

@markmcd Ready to merge then

PR #285, opened 2024-09-12
Improving caching notebooks

Giom-V:


quickstarts/Caching.ipynb
```diff
@@ -55,6 +55,15 @@
         "This notebook introduces context caching with the Gemini API and provides examples of interacting with the Apollo 11 transcript using the Python SDK. For a more comprehensive look, check out [the caching guide](https://ai.google.dev/gemini-api/docs/caching?lang=python)."
```
markmcd, line 55: <p>Try not to include numbers here. "Show, don't tell" is the philosophy we use. If anything changes, this becomes out of date and makes it hard to do things like update the model easily.</p><p> </p><p>For pricing especially, I would avoid numbers unless they are truly immutable. Pricing is charged by the hour, so there's no guarantee that this is any cheaper.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google-gemini/cookbook/pull/285/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='25'/>

Giom-V: In this case I prefer to keep it since it's the value added of the feature.

PR #288, opened 2024-09-12
Accept if a file is linked in it's folder's README

Giom-V:
* Limit the check to new .ipynb files
* When a file is in a sub-folder, accept it if there's a link to the file in a sub-folder's README


.github/workflows/new_examples_links_in_table_of_content.yml
```diff
@@ -15,30 +15,42 @@ jobs:
           git diff --name-status origin/main HEAD | grep ^A | cut -f2- > new_files.txt
           echo "Added Files:"
           cat new_files.txt  | sed 's/^/ - /'
-      - name: Check for new example files
+      - name: Check for new .ipynb files in the example folder
```
markmcd, line 18: Can you use something like [this](https://github.com/google-gemini/generative-ai-python/blob/4f42118d3d4daf9addc1c18f7d6895f7cd74087d/.github/workflows/samples.yaml#L17) here instead?

markmcd (approved):

.github/workflows/new_examples_links_in_table_of_content.yml
Giom-V: Good idea. I'll try it.

Giom-V: I've updated the script to use tj-actions/changed-files.

I've also tried to use a job-level FOLDER variable so we could keep the rest of the script the same when adding it for the quickstart folder but I did not manage to make it work. If you're knowledgeable about it I'd love some pointers.

markmcd (approved):

PR #290, opened 2024-09-13
Internal links as relative one

Giom-V:
* Fixing internal links in latest modified notebooks
* Updating template with relative links and instructions regarding links 
* Adding a "Next Steps" section in the template


Giom-V:
The only lint fail in related to the template, which I think is OK to have lint violations.

markmcd (approved):

PR #291, opened 2024-09-16
Add multimodal RAG tutorial

shilpakancharla:


Giom-V:
@shilpakancharla I think this PR still make sense, do you think you could port it to the new SDK?

Issue #201, opened 2024-06-08
Run on Multiple Images


sauravsinhaa:
### Description of the feature request:

I cant run on multiple images as i am trying to run it on many images and store there output but it is saying 
{
"image": "800040_2.pdf_4.jpg",
"description": "I am sorry, I cannot access or process any external files, including images. Therefore, I am unable to extract information from the provided image. \n\nTo assist you, I need the text content of the menu itself. If you can copy and paste the text here, I can help you process it and return the data in the JSON format you requested. \n"
}

### What problem are you trying to solve with this feature?

But when i run on single images it works  please let me know what i have done wrong

### Any other information you'd like to share?

import os
import json
import google.generativeai as genai
import google.generativeai as genai
from google.colab import userdata
GOOGLE_API_KEY=userdata.get('secretName')
genai.configure(api_key=GOOGLE_API_KEY)
model = genai.GenerativeModel('gemini-1.5-flash')

image_directory = '/content/unzipped_files/New folder/'

for filename in os.listdir(image_directory):
    if filename.lower().endswith(('.png', '.jpg', '.jpeg')):
        img_path = os.path.join(image_directory, filename)

        # Define your prompt
        prompt = """This image contains the restaurant menu we want to extract the dish name, section , price, description. section will be available at top only if not give blank section. Return output in json format:
                        {section: section , dish name: dish name, price:price, description:description}"""

        # Generate content using the model
        response = model.generate_content([prompt, img_path])  # Pass prompt and image path

        # Create a dictionary for this image
        image_data = {
            'image': filename,
            'description': response.text  # Access the text from the response
        }

        # Save the dictionary as a JSON file
        json_filename = os.path.splitext(filename)[0] + '.json'  # Remove extension and add .json
        json_path = os.path.join(image_directory, json_filename)
        with open(json_path, 'w') as f:
            json.dump(image_data, f, indent=4)

print("JSON files created successfully!")

singhniraj08:
@sauravsinhaa, Thank you reporting this issue.
This repository is for issues related to Cookbook guides and examples for Gemini API. The output is generated by the Gemini API and for issues related to Gemini API, we would suggest you to use "Send Feedback" option in [Gemini docs](https://ai.google.dev/gemini-api/docs/api-key). Ref: Screenshot below. You can also post this issue on [Discourse forum](https://discuss.ai.google.dev/).

[...]

PR #297, opened 2024-09-26
Fixes for Run on multiple Images

manojssmk:
This notebook demonstrates how to prompt multiple images and obtain a summary for each image in JSON format using threading. 

Fix : #201


Giom-V:
Hello @manojssmk,

Thanks for the PR. I'll have a look in details this week and I'll likely add a couple of things, like a "what's next" section at the end (cf. template).

In the meantime, could you try to fix the lint issues? It's mainly changing all the "we"s into "you"s.

Thanks!

markmcd:
A couple of quick drive-by comments from me:

1. Please make sure you comply with the licenses when distributing third-party content (the images)
2. Run pyink & nbfmt on the notebook to keep the code style-consistent 
3. I'm a little concerned that the code is "too smart" - does it need to demonstrate batching images into single prompts?
   * I would either put one image per request and a simple thread pool,
   * Or can you just put them all in one prompt? If performance is the point, then this is only 1 request, and it could show off the 2M context nicely.
   * Either way, it would help to understand what you are trying to demonstrate.

PR #301, opened 2024-10-04
Define the websiteCopyPrompt variable

Hamza-nabil:
Define the websiteCopyPrompt variable  in  `examples/Market_a_Jet_Backpack.ipynb` notebook


markmcd (approved):

markmcd:
We really should use `lower_case_variables` but that's not your fault :)

Thanks for the contrib!

PR #303, opened 2024-10-07
Simplify AIS template

MarkDaoust:
This is about the simplest I can make it. The many ways that AIS may send a file are the source of some of the complexity. A "no-files" version could simplify things a lot.

We loose three things if we do that:

1. There's no "chat" (but we could split `chat` to it's own template).
2. It's harder for users to tell what went wrong if something breaks.
3. I removed the "show the conversation" code. 
   - It's nice that it lets the user see what they're sending.
   - But it is long/complex.

WDYT?


templates/aistudio_gemini_prompt_freeform_nofiles.ipynb
```diff
@@ -0,0 +1,171 @@
...
+      "outputs": [],
+      "source": [
+        "!pip install -U -q \"google-generativeai>=0.8.2\""
+      ]
+    },
+    {
+      "cell_type": "code",
+      "execution_count": 11,
+      "metadata": {
+        "id": "kWIuwKG2_oWE"
```
MarkDaoust, line 58:
```suggestion
        "cellView": "form",
        "id": "kWIuwKG2_oWE"
```

MarkDaoust:
I think this is ready to submit.

markmcd (approved):

LORDWORLD10:
Pekný deň prajem , rozmýšľam do akej kategórie vás zaradíme a kto ste čo
ste a čo bude náš spoločný záujem.?

Dňa ut, 5. nov 2024, 2:57 Mark Daoust ***@***.***> napísal(a):

PR #310, opened 2024-10-18
Add an async quickstart

markmcd:
This is a Python-only quickstart that shows how to use the SDK's `async` API.

I've included two examples, one as the "simplest possible" case that assumes the files are available locally and a more complex example that schedules in parallel and includes a download step using a shared HTTP session.


MarkDaoust (approved):
Awsome!

PR #318, opened 2024-11-08
Add Project IDX support

rodydavis:
Support for Project IDX:
https://idx.dev

TLDR: Opens the cookbook in IDX as a template with a full VM and filesystem.

Test the template here:
<a href="https://idx.google.com/new?template=https%3A%2F%2Fgithub.com%2Frodydavis%2Fcookbook">
  <picture>
    <source
      media="(prefers-color-scheme: dark)"
      srcset="https://cdn.idx.dev/btn/try_light_32.svg">
    <source
      media="(prefers-color-scheme: light)"
      srcset="https://cdn.idx.dev/btn/try_dark_32.svg">
    Try in IDX
  </picture>
</a>


requirements.txt
Giom-V, line 1: Would there be a way to move those files out of the main folder? Ideally I would prefer to have it as clean as possible and not risk a for a beginning user to run the requirements for nothing.

README.md
```diff
@@ -44,7 +44,28 @@ The Gemini API is a REST API. You can call the API using a command line tool lik
...
+    <img
+      height="32"
+      alt="Try in IDX"
+      src="https://cdn.idx.dev/btn/try_purple_32.svg">
+  </picture>
+</a>
+
+To run any notebook in [IDX](https://idx.dev) replace the TODO in `.idx/dev.nix` with your Gemini API key by:
+- Selecting "Add Gemini API" in the "Project IDX" panel in the sidebar
+- Or by visiting https://g.co/ai/idxGetGeminiKey
```
Giom-V, line 68: I guess you want to track how many keys were created through IDX but I would prefer if we do not use different urls for the same website on the same page so I'd use https://aistudio.google.com/app/apikey

scripts/api-key.py
```diff
@@ -0,0 +1,35 @@
+import os
+
+def update_ipynb_files(directory):
```
Giom-V, line 3: I think you will also need to update all the internal links, otherwise the links in the "what's next" sections of each notebook will open the github instead of the corresponding notebook in IDX, breaking the learning path.

README.md
rodydavis: Yep that was an old url, happy to update! We also have a seamless flow in IDX to create API keys without leaving the editor

requirements.txt
rodydavis: I am not too familiar with how python uses that file, just thought it was a convention. If I can move it then I agree that would be cleaner!

scripts/api-key.py
rodydavis: Sounds good! Can colab use the system ENV if the keys are defined in the sidebar? Ideally the code would not have to change either.

But also agreed about relative links

PR #349, opened 2024-12-13
Update live_api_starter.ipynb

logankilpatrick:


random-forests (approved):

Giom-V:
We wrote that in all the notebooks, why remove it only in this one?

PR #351, opened 2024-12-14
Update README.md

logankilpatrick:


Giom-V (approved):
I think we also need to add links to the new SDK. Either do it or merge that CL and I'll do it on Monday.

PR #353, opened 2024-12-15
Create async queues within async loop

clementfarabet:
Code doesn't run as is / crashes with queues created outside of asyncio.run().

### DIFF
--- a/gemini-2/live_api_starter.py
+++ b/gemini-2/live_api_starter.py
@@ -55,9 +55,9 @@ pya = pyaudio.PyAudio()
 
 class AudioLoop:
     def __init__(self):
-        self.audio_in_queue = asyncio.Queue()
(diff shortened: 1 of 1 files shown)


clementfarabet:
For context, error without this change:

```
RuntimeError: Task <Task pending name='Task-8' coro=<AudioLoop.send_frames() running at /Users/clmt/workspace/cookbook/gemini-2/live_api_starter.py:107> cb=[_TaskGroup._on_task_done(), AudioLoop.run.<locals>.check_error() at /Users/clmt/workspace/cookbook/gemini-2/live_api_starter.py:194]> got Future <Future pending> attached to a different loop
```

MarkDaoust (approved):
Thanks, it looks like this depends on pythion version and we hadn't tested some of the older ones.

clementfarabet:
Thanks!

PR #369, opened 2024-12-18
Update links in get_started.ipynb

shilpakancharla:
Updated to use Gemini API links instead of Vertex AI links where needed


Giom-V (approved):
Good catch!

PR #375, opened 2024-12-19
added the AudioLoop module

dtiberio:
added the AudioLoop module that implements the code from live_api_starter.py as a Python module that can be imported into other apps.


MarkDaoust (changes requested):
Oh, I hadn't see this one. 

In the description can you add high-level bulletted list of wha changes you're trying to make.

I made a couple of similar changes in https://github.com/google-gemini/cookbook/pull/377, sorry I didn't see this first. 

I don't think we want to duplicate this file.  If we want to move in this direction we should make the changes in the original file, instead of forking.
Also, any change here should be mirrored in the websockets/ version.

PR #376, opened 2024-12-19
Use `thought` flag.

MarkDaoust:


Giom-V:
<p>As you'll see in my big comment, I have mixed feeling about using a function to display the output of the calls. I think we should keep the examples self-understandable as much as possible when it's just about a couple easy to understand lines.</p><p> </p><p>That said, I'm also concerned that the tag is only available in v1alpha as someone jumping directly to the examples won't understand why it's not working for them, so I'm not sure it's a good idea to have our examples be reliant on the alpha API.</p><p> </p><p>Personally I would only add streaming (and maybe chat) for the time being and wait for v1beta to display the thought tag before updating the rest.</p>

Issue #383, opened 2024-12-22
Gemini 2.0 Flash spatial reasoning severly degraded with Image size 1024


fayezsalka:
### Description of the bug:

I was trying to reproduce results I get with the spatial reasoning applet found in the AI studio, using the spatial reasoning notebook. However, I noticed for the same prompt and same image, applet produces far better results with 2.0 Flash. 

Upon further investigation, it seems that in the notebook the image is resized to 1024, but in the applet the image is resized to 640. Using images with size of 1024 significantly degrades the quality of the boxes generated. E.g a lot of the bounding boxes are misplaced and detected incorrectly. Changing the sizing to use maximum 640 for width in the notebook fixes the issue.

Any intuition on why image sizes 1024 degardes the spatial reasoning?

fayezsalka:
Here is the comparison between 640 and 1024 image size:

**640:**
![image](https://github.com/user-attachments/assets/7507d72c-6bfd-4a3f-a856-d18165d65819)

**1024:**
![image](https://github.com/user-attachments/assets/01578552-ebf3-4c4f-a180-2be16632a3db)

The degraded quality is not specific to this image / prompt. Same issue happens with all the examples found in spatial reasoning. Cupcakes, Fox, etc

Giom-V:
Thanks for reporting that behavior! I'll check with the Deepmind researchers but I'd say that it's because there is too much information when the image is not resized.

lucianommartins:
Hi @fayezsalka, 

Can you share a colab with the experiment you did? I tried reproducing the misbehavior you are facing using the following image but it consistently worked with `resize()` set to `[1024, 768, 640, 512, 256]`:

![image](https://github.com/user-attachments/assets/f34192bf-9c97-4304-b0e7-b679da2b70e9)

Thanks!!

Giom-V:
@fayezsalka I've updated the notebook to resize the images to 640 in #396.

PR #396, opened 2025-01-06
Resizing images to 640

Giom-V:
Resizing most images to 640 for better results as discussed in #383 and to have the same behavior as in the AI Studio demo.


MarkDaoust (changes requested):
The variable naming isn't great here: `Image`, `image`, `img` and `im`. 

I think this:

```
image = "Socks.jpg"

# Load and resize image
img = Image.open(BytesIO(open(image, "rb").read()))
im = Image.open(image).resize((640, int(640 * img.size[1] / img.size[0])), Image.Resampling.LANCZOS)
```


```
image_path = 'cupcakes.jpeg'
im = Image.open(image_path)
im.thumbnail([640,640], Image.Resampling.LANCZOS)
```

Giom-V:
Good point, I think it's a remnant of the multiple rewritings. I'll update accordingly.

Giom-V:
> im.thumbnail([640,640])

One thing though is that this makes the image square, which is why I have to be more creative to keep its format.

Giom-V:
I remember why it was like this, it's so that we don't have to reload the initial image again when running the script (since the bounding box script overwrites them). But I think it's ok in this case.

Giom-V:
Interestingly, the code you proposed seems to have an impact on performances (like the spill example works 100% of the time with `resize` and doesn't work 100% of the time with `thumbnail`, so I'm guessing it must not be resizing the exact same way and some details are lost.

I'll stick with what works.

MarkDaoust (approved):
> I'll stick with what works.

Yes. Thanks.

PR #399, opened 2025-01-09
New version requires `input=` keyword for arg.

MarkDaoust:


Giom-V (approved):
Approved but shouldn't you also update live_api_tool_use.ipynb as well?

PR #403, opened 2025-01-12
Switch gemini-pro to gemini-1.5-flash

logankilpatrick:


evansenter (approved):
LGTM, thanks for the quick fix!

PR #404, opened 2025-01-13
Add license header checking app

markmcd:
Almost the same config as the generative-ai-docs repo: https://github.com/google/generative-ai-docs/blob/main/.github/header-checker-lint.yml


Giom-V (dismissed):
Approved, but have you run it on the existing .py files? Do we need to update their licences?

markmcd:
> Approved, but have you run it on the existing .py files? Do we need to update their licences?

Yeah, they will need to be updated too. Typically I would do the "prevent backsliding" step before the "update the world" step, since they're independent, but it looks like there's only the file-api samples so I'll update it here.

Giom-V (approved):

PR #406, opened 2025-01-15
Fixing internal links in github

Giom-V:
The links will still not go to the right parts but at least it will stay in the same file.


markmcd (approved):

markmcd:
how annoying

Issue #386, opened 2024-12-27
I get Error with CONFIG = {"generation_config": {"response_modalities": ["AUDIO","TEXT"]}} in gemini-2/live_api_starter.py


abdul7235:
### Description of the bug:

From the documentation: 

https://ai.google.dev/api/multimodal-live

I believe I can get response in multiple modalilties, but running the above CONFIG in my code I get following error:

websockets.exceptions.ConnectionClosedError: received 1007 (invalid frame payload data) Request trace id: a0bb7a2dd8834b47, [ORIGINAL ERROR] generic::invalid_argument: Error in program Instantiation for language; then sent 1007 (invalid frame payload data) Request trace id: a0bb7a2dd8834b47, [ORIGINAL ERROR] generic::invalid_argument: Error in program Instantiation for language

Why am I getting this error?

### Actual vs expected behavior:

Is it possible to have response in Audio and Text Simaltaneoulsy? 

If yes, please help me sort it out.

If no, then for good's sake it must be mentioned clearly in the documentation!

### Any other information you'd like to share?

I had appreciate if google and GCP had properly organized and clear documentation for their services, APIs and SDKs. It is such a horrible experience integrating google's servicses due to documentation being horribly scattered and vague.

LarsDu:
I'm getting the same error. It would be nice to get both text and audio at the same time. This is particularly useful for generating dialogues for things like games...

Giom-V:
Hello @abdul7235 and @LarsDu,

At the moment, multimodalities is not available publically. You can only get Audio using the live APIs, and text using the "classic" ones.

I'll see what can be done to make that clear in the documentation.

abdul7235:
@Giom-V 

Will multimodal output be available in the coming months? 

Also, could you guide me on how to gain access to the non-public multimodal API?

kshitij01042002:
Is it possible to get the Audio and function calling? @Giom-V

simix:
this is when you sent unsafe prompts  
im looking how to block it , i tried to use this and doesnt work 

self.safety_settings = [
            {
                "category": "HARM_CATEGORY_DANGEROUS",
                "threshold": "BLOCK_NONE",
            },
            {
                "category": "HARM_CATEGORY_HARASSMENT",
                "threshold": "BLOCK_NONE",
            },
            {
                "category": "HARM_CATEGORY_HATE_SPEECH",
                "threshold": "BLOCK_NONE",
            },
            {
[...]

PR #416, opened 2025-01-23
fix: Add validation and documentation for response modalities

paulelliotco:
This PR addresses issue #386 by adding input validation and clear documentation for response modalities in the live_api_starter.py file. 
The changes include:
- Added validate_config() function to check for invalid or unsupported response modalities.
- Improved error handling in the run() method to provide clear feedback for invalid configurations.
- Updated documentation comments to clarify the limitations of response modalities.

### DIFF
--- a/gemini-2/live_api_starter.py
+++ b/gemini-2/live_api_starter.py
@@ -83,6 +83,32 @@ CONFIG = {"generation_config": {"response_modalities": ["AUDIO"]}}
 
 pya = pyaudio.PyAudio()
 
+def validate_config(config):
+    """
+    Validate the generation config, specifically checking response modalities.
+    
+    Note: Currently only one response modality is supported at a time.
+    AUDIO is supported in Live API, TEXT is supported in classic API.
+    Multimodal output (AUDIO + TEXT) is not yet publicly available.
+    """
+    if "generation_config" in config:
+        modalities = config["generation_config"].get("response_modalities", [])
+        if len(modalities) > 1:
+            raise ValueError(
+                "Currently only one response modality is supported at a time. "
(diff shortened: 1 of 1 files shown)


markmcd:
I like your idea here - however we don't typically show client-side validation of configuration for a few reasons.

1) the truth is ultimately on the server-side, so client-side validation code can become incorrect over time (this creates a maintenance burden for us here in the cookbook, or broken samples).
2) data validation is often different per-user, so in this example, a user (either a public user in the future, or a preview user now) would see a false validation error.
3) some amount of validation can be done by using objects (i.e. proto derived types) instead of python dicts - e.g. `types.LiveConnectConfig(response_modalities=[types.Modality.AUDIO])` (where `types` is `google.genai.types`). it doesn't apply to this specific case, but it's a general reason why we don't include validation in sample code.

So while `['AUDIO']` and `['TEXT']` are the only valid values right now, I don't think it's worth the extra validation code.

If you think that this particular tool is somewhere that people will get caught up, perhaps add a comment along the lines of `# While Gemini 2.0 Flash is in experimental preview mode, only one of AUDIO or TEXT may be passed here.`

Or if you have a better suggestion that would reach more users (e.g in the docs on ai.google.dev somewhere), please share.

I'll close this now but if you want to send a PR to add a comment, feel free.

paulelliotco:
Thank you Mark for the detailed feedback! 

I completely understand the reasoning behind avoiding client-side validation. I'll add a comment to clarify the limitation of response modalities in the `live_api_starter.py` file and  submit a PR with this update shortly.

AlipMagical:
Hello I am optioning all out wait okay..I will show you all okay

PR #429, opened 2025-01-31
Update Get_started_thinking.ipynb

gmKeshari:
Removed  "`config={'thinking_config': {'include_thoughts': True}},\n`", from the code script because it gives 400 error as `include_thoughts` is not supported with "gemini-2.0-flash-exp".


Giom-V:
Thanks for the submission, this was already done in a parallel one (https://github.com/google-gemini/cookbook/pull/426) so I'm closing yours.

PR #431, opened 2025-02-04
Use non-experimental model name for 2.0

markmcd:
I updated the flash-exp model references to drop the `-exp` suffix. Also switched each of these to use a model selector where it wasn't already configured.

I haven't added any extra models to notebooks with the selector as I figure they should be tested, and I didn't do that, but we can do it in future updates if we wish.

And included `isTemplate` so users (or us?) can link to specific models in Colab.


Giom-V:
I think there's still the open question about "isTemplate: true", and you're still adding dropdowns in some place I think I forgot, so up to you to decide if you want to bother with the merge.

PR #455, opened 2025-02-12
Update Audio notebook

patrickloeber:
Update audio notebook to use new SDK and add more info as on the [docs page](https://ai.google.dev/gemini-api/docs/audio?lang=python)


philschmid (dismissed):
LGTM. I would maybe just remove the install output in the first cell. Thanks for updating to the new sdk and model.

philschmid (approved):
GTLM

PR #468, opened 2025-02-20
Adding a Imagen example to illustrate a book

Giom-V:


philschmid (approved):
Great Examples! LGTM. But i think we should remove the images from the notebooks and add rendered ones via markdown hosted on the bucket. Each Image is ~1MB which is a lot.

PR #470, opened 2025-02-20
fix cache in getting started notebook

patrickloeber:
Code example was broken


philschmid (approved):
LGTM! Maye we can add a note/comment that this currently works with 1.5 and not 2.0 but we are working on it.

Giom-V:
> LGTM! Maye we can add a note/comment that this currently works with 1.5 and not 2.0 but we are working on it. 

We already say "It is not yet available in Gemini 2.0 because it is an experimental model." But we could be clearer.

patrickloeber:
yes, I added one more sentence about supported versions

philschmid (approved):

Giom-V:
Thank you both!

PR #495, opened 2025-03-04
migrate: Guess_the_shape.ipynb to new SDK

haroon10725:
@Giom-V @markmcd Can you please review it.

### DIFF
--- a/examples/Guess_the_shape.ipynb
+++ b/examples/Guess_the_shape.ipynb
@@ -63,7 +63,7 @@
       },
       "outputs": [],
       "source": [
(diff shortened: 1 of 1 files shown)


markmcd (approved):
I've made a couple of fixes:

1. Include the SDK version when it's loaded (this helps users understand what version this was run with if something changes)
2. Fixed the wrong `generate_content` call
3. Included the updated output from the model (this is important if the model ever changes - we need to make sure the output matches the model)

markmcd:
Thanks for the contrib!

PR #484, opened 2025-03-01
Update package to `google-genai` in `examples/Anomaly_detection_with_embeddings.ipynb`

Faakhir30:
Refers #446 

- [x] Updated `google-generativeai` to use `google-genai` package instead.
- [x] Updated outputs and tested locally.
- [x] Process 100 rows in 1 batch request for getting embeddings. (100 is the max limit `BatchEmbedContentsRequest.requests: at most 100 requests can be in one batch`) .The original issue didn't require this optimization, but it was needed since free gemini API credits are insufficient for 600 requests. Now, it can be done in 6 requests.
- [x] formatted with `pyink examples/Anomaly_detection_with_embeddings.ipynb`

Would update the remaining notebooks (one at a time) once this one is merged successfully. thanks

PS: This is my first `ipynb` oss contribution, so not expecting this to be perfect in the first attempt, but I'm willing to learn n update PR :)

Issue #446, opened 2025-02-07
Update everything to use google-genai SDK


markmcd:
### Description of the feature request:

[Lots of notebooks are using the `google-generativeai` SDK](https://github.com/search?q=repo%3Agoogle-gemini%2Fcookbook%20%22import%20google.generativeai%22&type=code), but now that [`google-genai`](https://pypi.org/project/google-genai) is at 1.0, we should be using that.

e.g. `pip install google-generativeai` needs to be replaced with `pip install google-genai` and then relevant code and text needs to be updated.

There's lots to do right now, but we can do every notebook independently so this would be a great one to open to community contributions. If you want to work on one, check the open PR list to see if someone has already done it, then just send a PR directly (but stick to one notebook per PR).

mitrick2:
Are there plans to create a google-genai SDK for JS?

Faakhir30:
Updating `Anomaly_detection_with_embeddings.ipynb`. Informed to avoid duplicate efforts by other contributors, thanks

Faakhir30:
> Are there plans to create a google-genai SDK for JS?

I guess we already have it https://github.com/google-gemini/generative-ai-js, currently at `google/generative-ai@0.8.0` at the time of comment.

charantek1styearbtech:
@markmcd need help in accepting Google contributor licence agreement there is always an server error.
Any suggestions?

ROHANDWIVEDI2005:
> [@markmcd](https://github.com/markmcd) need help in accepting Google contributor licence agreement there is always an server error. Any suggestions?

I also have the same issue.

Giom-V:
> Are there plans to create a google-genai SDK for JS?

@mitrick2 A new typescript unified SDK is on its way.

Giom-V:
@charantek1styearbtech @ROHANDWIVEDI2005 @Faakhir30 As for the CLA issue, it's being investigated internally. I'll update this thread when it will be fixed.

Thanks you all for your contributions!

Giom-V:
Hey all,

Thanks a lot for all your submissions! This is greatly helping us.

I've fallen sick and I have to prepare 4 talks for next week ([here](https://cloudonair.withgoogle.com/events/build-with-gemini?talk=talk-1-pt), if you are in Paris) so I couldn't find the time to review all the Pull requests. But rest assured that I'm not forgetting you and I'll catch up with everything you submitted next week.

KSruthiVel:
[...]

PR #497, opened 2025-03-04
Update Analyze_a_Historic_Event_Recognition to use genai module

Faakhir30:
Refers #446
- [x] Updated notebook to use `genai` module
- [x] Refactoed imports
- [ ] Updated Readme Links

Readme links of Video analysis notebooks are updated once in #499 , so not needed over here, would implement here again if required.

Learings from #484 are implemented over here, hopefully would require less review effort :)


Giom-V (approved):
Thanks for the submission and the help getting the examples to use the new SDK!

PR #506, opened 2025-03-05
Update `Sentiment_Analysis` notebook to use genai module

Faakhir30:
Refers #446

- Updated module `google-generativeai` to use `google-genai>=1.0.0`
- Refactored imports
- Docs update was not required IMO

### DIFF
--- a/examples/json_capabilities/Sentiment_Analysis.ipynb
+++ b/examples/json_capabilities/Sentiment_Analysis.ipynb
@@ -73,27 +73,13 @@
           "name": "stdout",
           "output_type": "stream",
           "text": [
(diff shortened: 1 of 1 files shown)


markmcd:
Thanks for the update on this one - I like the work you've done and am happy to merge it but I think this doc needs more work. 

Just to be clear, your work is great, we can track this separately. But if you are interested it'd be great to work on this.

Here's what I think needs to be done:
* The schema should be moved out of the prompt and should use `response_schema` instead (so we use a real `TypedDict` or pydantic class)
* We should use categorical outputs in the schema (like enums) instead of float outputs. The floats are not going to be calibrated probabilities, so we shouldn't represent them as such. Instead each field could be one of `NONE`, `WEAK`, `STRONG`.

If you like, we can just merge this and deal with these tasks in a separate issue. LMK what you'd like to do.

markmcd (approved):

markmcd (approved):

PR #508, opened 2025-03-05
Update Text_Summerization notebook to use genai module

Faakhir30:
Refers #446

- Updated module `google-generativeai` to use `google-genai>=1.0.0`
- Refactored imports
- Docs update was not required IMO

PR #507, opened 2025-03-05
Update Text_Classification notebook to use genai module

Faakhir30:
Refers #446

- Updated module `google-generativeai` to use `google-genai>=1.0.0`
- Refactored imports
- Docs update was not required IMO

PR #509, opened 2025-03-05
Update Entity_extraction_JSON notebook to use genai module

Faakhir30:
Refers #446

- Updated module `google-generativeai` to use `google-genai>=1.0.0`
- Refactored imports
- Docs update was not required IMO

Following PRs completes the refactor and update of `examples/json-capabilities/*`:
- Entity_Extraction_JSON.ipynb (This PR)
- Sentiment_Analysis.ipynb #506
- Text_Summarization.ipynb #508
- Text_Classification.ipynb #507

Also, I'm following the instructions in the above-referenced issue to follow 1 notebook max per PR, but I think there are some small notebooks like the above four that could have been a single PR since they are related a bit and are not that complex/lengthy. Also have gained experience in updating notebooks :) 
So would like to make a single PR(ofc if it makes sense) from now on if allowed by maintainers. thanks

### DIFF
--- a/examples/json_capabilities/Entity_Extraction_JSON.ipynb
+++ b/examples/json_capabilities/Entity_Extraction_JSON.ipynb
@@ -64,28 +64,13 @@
     },
     {
       "cell_type": "code",
-      "execution_count": 8,
+      "execution_count": 5,
       "metadata": {
         "id": "Ne-3gnXqR0hI"
       },
       "outputs": [],
       "source": [
-        "!pip install -U -q \"google-generativeai>=0.7.2\""
-      ]
-    },
-    {
-      "cell_type": "code",
-      "execution_count": 9,
-      "metadata": {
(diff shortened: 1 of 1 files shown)


Giom-V (approved):
Thanks again @Faakhir30!

Giom-V:
> Also, I'm following the instructions in the above-referenced issue to follow 1 notebook max per PR, but I think there are some small notebooks like the above four that could have been a single PR since they are related a bit and are not that complex/lengthy. Also have gained experience in updating notebooks :) So would like to make a single PR(ofc if it makes sense) from now on if allowed by maintainers. thanks

@Faakhir30 I think it's better for everyone's visibility to keep one PR per notebook so that it limits the chances that two contributors work on the same thing. That said, if you are clear in the RP name, you can mix multiple notebooks in one.

PR #510, opened 2025-03-05
Improve error handling and robustness in gradio_audio

Jay-Lokhande:


markmcd (changes requested):
It'd help to understand more about these changes.  Are you changing things to handle specific cases that you encountered? Or is this general code cleanup?

To set expectations, this is sample code, so we tend to prefer the code be clear and readable over production-hardened, even if that means it can't handle every possible error. Important or common errors should absolutely be addressed though.

PR #521, opened 2025-03-06
updated Embeddings.ipynb to use latest python sdk

ROHANDWIVEDI2005:
Can anyone tell me if the syntax for **taskType** of the  **models.embedContent method** has changed in the new python sdk.
I cheked the embeddings docs for it.
Screenshot 2025-03-06 103814
does this mean **taskType** can only be used for **embedding-001** because in the embeddings notebook is using it with **embedding-004**.


quickstarts/Embeddings.ipynb
```diff
@@ -72,7 +72,7 @@
       },
```
markmcd, line 72: <pre class="ql-syntax" spellcheck="false">Line #2.    result = client.models.embed_content(model="text-embedding-004", content=text)
</pre><p> </p><p>Please run the colab and make sure everything is correct.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google-gemini/cookbook/pull/521/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

markmcd, line 72: <pre class="ql-syntax" spellcheck="false">Line #5.    print(str(result.embeddings)[:50], '... TRIMMED]')
</pre><p> </p><p>This isn't quite the same output as before... Can you use <code>result.embeddings[0].values</code> instead of <code>result.rembeddings</code> here?  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google-gemini/cookbook/pull/521/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

markmcd, line 72: <pre class="ql-syntax" spellcheck="false">Line #1.    print(len(result.embeddings)) # The embeddings have 768 dimensions
</pre><p> </p><p>This is also incorrect.</p><p> </p><p>Since this is batching/unbatching, it might be clearer to use lists as input and unpack the results when they're generated. e.g.</p><p> </p>

<pre class="ql-syntax" spellcheck="false">texts = ["Hello world"]

result = client.models.embed_content(model="text-embedding-004", contents=texts)

[embedding] = result.embeddings
print(len(embedding.values))
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google-gemini/cookbook/pull/521/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

markmcd, line 72: <p>This is broken too. Please run this &amp; ensure it executes and that the outputs are correct.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google-gemini/cookbook/pull/521/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

markmcd (changes requested):
See reviewnb comments

markmcd:
Thanks for the contibution!

> does this mean taskType can only be used for embedding-001 because in the embeddings notebook is using it with embedding-004.

No that doc comment looks wrong. It should work with the text004 model too (as seen in this notebook)

quickstarts/Embeddings.ipynb
ROHANDWIVEDI2005: <p>done</p>

ROHANDWIVEDI2005: <p>yes, i updated this as well</p>

markmcd:
> Another thing that I have noticed recently is that in the generative-ai-python most of the samples are using the google-genrativeai SDK instead of the new google-genai SDK . I checked the issues section and apparently there is no issue open about it?

That repo **is** the `google-generativeai` SDK, so the samples should match. The new samples are in [google-gemini/api-examples](https://github.com/google-gemini/api-examples/tree/main/python), and these should be linked from the [API reference](https://ai.google.dev/api/generate-content).

Please let us know if you saw somewhere that links to that repo specifically for the context of `google-genai` examples - if it is causing confusion we should fix it.

markmcd (dismissed):

markmcd (approved):

markmcd:
Looks like nblint is broken (FYI @Giom-V). merging anyway since it works when I run it locally with an older version.

ROHANDWIVEDI2005:
> > Another thing that I have noticed recently is that in the generative-ai-python most of the samples are using the google-genrativeai SDK instead of the new google-genai SDK . I checked the issues section and apparently there is no issue open about it?
> 
> That repo **is** the `google-generativeai` SDK, so the samples should match. The new samples are in [google-gemini/api-examples](https://github.com/google-gemini/api-examples/tree/main/python), and these should be linked from the [API reference](https://ai.google.dev/api/generate-content).
> 
> Please let us know if you saw somewhere that links to that repo specifically for the context of `google-genai` examples - if it is causing confusion we should fix it.


[ai.google.dev-embeddings](https://ai.google.dev/gemini-api/docs/embeddings) page uses the new SDK but the `use cases` for them link to [generative-ai-docs](https://github.com/google/generative-ai-docs) which uses the old SDK this could cause confusion.

Screenshot 2025-03-07 111740
Screenshot 2025-03-07 111817

PR #527, opened 2025-03-06
Update Get_started_LiveAPI.py

MarkDaoust:
This is broken.


markmcd (approved):

pzoltowski:
is there any up to date docs where I can see where stuff got deprecated how to migrate? I previously used CONFIG with 'system_instructions' and now not sure how to migrate. Some deprecation info in changelog would be also useful.

PR #536, opened 2025-03-08
update Agents_function_calling_barista_Bot

AadiSharma49:
**Fixed & Improved Chat with Barista Bot**

1.  Prevents Infinite Loop: Allows the user to type "exit", "quit", or "done" to end.
2.  Handles Errors Properly: Checks if response is valid before displaying it.
3.  User-Friendly Formatting: Uses better spacing, emojis, and Markdown for better readability.
4.  Ensures a Proper Final Order Summary.


AadiSharma49:
@Giom-V Could you please review and approve this PR? Thanks!

AadiSharma49 (commented):
check it !

markmcd:
Can you please rebase your changes to see the latest updates?

examples/Agents_Function_Calling_Barista_Bot.ipynb
```diff
@@ -118,64 +118,80 @@
...
+        "    if not order:\n",
+        "        print('  (no items)')\n",
+        "    else:\n",
+        "        for drink, modifiers in order:\n",
+        "            print(f'  {drink}')\n",
+        "            if modifiers:\n",
+        "                print(f'   - {\", \".join(modifiers)}')\n",
         "\n",
-        "  return input('Is this correct? ')\n",
+        "    return input('Is this correct? (yes/no): ').strip().lower()\n",
```
markmcd, line 180: This isn't necessary - part of the "magic" here is that the user can say whatever they want and the model will interpret it.

examples/Agents_Function_Calling_Barista_Bot.ipynb
```diff
@@ -327,22 +360,79 @@
       },
       "outputs": [],
       "source": [
+        "from typing import Iterable\n",
```
markmcd, line 363: Did you mean to leave an extra copy of the order system here?

examples/Agents_Function_Calling_Barista_Bot.ipynb
```diff
@@ -327,22 +360,79 @@
...
         "@retry.Retry(initial=30)\n",
         "def send_message(message):\n",
-        "  return convo.send_message(message)\n",
-        "\n",
-        "\n",
-        "placed_order = []\n",
-        "order = []"
+        "    try:\n",
+        "        return convo.send_message(message)\n",
+        "    except Exception as e:\n",
```
markmcd, line 433: What kind of error are you handling here? It would be good to 1) be more explicit about which error it is and 2) if there's nothing in particular, just let them float up to the user - it keeps the code simpler and (since we know the reader is a developer) they should see the errors (with full stack trace) anyway.

One design intent I had with this example is that the run-loop is super simple.  I like the exiting strategy but if there's no need to add extra error handling then we can just omit it.

markmcd (changes requested):
Thanks for the updates. I've left a few comments, but first up please rebase to pick up the changes from the new SDK.

examples/Agents_Function_Calling_Barista_Bot.ipynb
AadiSharma49: Thanks for the feedback, Mark — that makes sense. I was initially catching the exception to prevent the app from crashing unexpectedly during send_message() in case of network issues or timeouts.

PR #540, opened 2025-03-09
Update Get_started_LiveAPI.ipynb

ajitashwathr10:
Fixed the links (redirecting to main branch) in Next steps

### DIFF
--- a/quickstarts/Get_started_LiveAPI.ipynb
+++ b/quickstarts/Get_started_LiveAPI.ipynb
@@ -664,7 +664,7 @@
         "- If you're interested in the low level details of using the websockets directly, see the [websocket version of this tutorial](../gemini-2/websockets/live_api_starter.ipynb).\n",
         "- Try the [Tool use in the live API tutorial](../quickstarts/Get_started_LiveAPI_tools.ipynb) for an walkthrough of Gemini-2's new tool use capabilities.\n",
         "- There is a [Streaming audio in Colab example](../gemini-2/websockets/live_api_streaming_in_colab.ipynb), but this is more of a **demo**, it's **not optimized for readability**.\n",
-        "- Other nice Gemini 2.0 examples can also be found in the [Cookbook](/google-gemini/cookbook/tree/main/gemini-2/), in particular the [video understanding](../quickstarts/Video_understanding.ipynb) and the [spatial understanding](../quickstarts/Spatial_understanding.ipynb) ones.\n"
+        "- Other nice Gemini 2.0 examples can also be found in the [Cookbook's 2.0 directory](https://github.com/google-gemini/cookbook/tree/main/gemini-2/), in particular the [video understanding](../quickstarts/Video_understanding.ipynb) and the [spatial understanding](../quickstarts/Spatial_understanding.ipynb) ones.\n"
       ]
     }
   ],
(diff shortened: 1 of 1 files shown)


ajitashwathr10:
I signed in CIA. What's next?

markmcd (changes requested):
If these are still necessary, can you make sure they are still relative links? This way the links behave as expected in Colab, Dev containers, etc.

ajitashwathr10:
Thanks for the review. But the issue is, I noticed that the current relative links are pointing to an outdated branch rather than the main branch. When users click these links, they're redirected to older versions of the content that may be outdated or inconsistent with the current documentation.

I changed them to absolute URLs to ensure they point to the current main branch content. This way, users will always see the most up-to-date version of the referenced files.

If we need to keep relative links for cross-platform compatibility (Colab, Dev containers, etc.), could you suggest the correct relative path structure that would point to the current branch? I'm happy to update the PR with proper links.

markmcd:
> Thanks for the review. But the issue is, I noticed that the current relative links are pointing to an outdated branch rather than the main branch. When users click these links, they're redirected to older versions of the content that may be outdated or inconsistent with the current documentation.

I think what you're seeing is how GitHub "pins" the version (to HEAD). If I load https://github.com/google-gemini/cookbook/blob/main/quickstarts/Get_started_LiveAPI.ipynb and look at the links at the end, the links use commit 2c04c8d13e3790180b7bdf4afcd2d9e219bb586b, which is the latest commit at HEAD.

I assume GitHub does this to ensure users are browsing a stable snapshot of the repository, and it prevents things like a delete or refactor commit from causing a 404 while they are browsing.

IMO this is fine, it's not going to be really old content unless someone keeps a tab open for ages. And importantly, it's consistent across GitHub.

ajitashwathr10:
I've reverted the changes back to use relative links as you suggested. I understand now that GitHub's "pinning" behavior is intentional and beneficial for ensuring a consistent browsing experience.

Thank you for explaining GitHub's approach with version pinning. It makes sense that this creates a more stable browsing experience while still maintaining the benefits of relative links for cross-platform compatibility.
Let me know if there's anything else I should change.

Giom-V (dismissed):

quickstarts/Get_started_LiveAPI.ipynb
```diff
@@ -664,7 +664,7 @@
         "- If you're interested in the low level details of using the websockets directly, see the [websocket version of this tutorial](../gemini-2/websockets/live_api_starter.ipynb).\n",
         "- Try the [Tool use in the live API tutorial](../quickstarts/Get_started_LiveAPI_tools.ipynb) for an walkthrough of Gemini-2's new tool use capabilities.\n",
         "- There is a [Streaming audio in Colab example](../gemini-2/websockets/live_api_streaming_in_colab.ipynb), but this is more of a **demo**, it's **not optimized for readability**.\n",
-        "- Other nice Gemini 2.0 examples can also be found in the [Cookbook](https://github.com/google-gemini/cookbook/tree/main/gemini-2/), in particular the [video understanding](../quickstarts/Video_understanding.ipynb) and the [spatial understanding](../quickstarts/Spatial_understanding.ipynb) ones.\n"
+        "- Other nice Gemini 2.0 examples can also be found in the [Cookbook's 2.0 directory](https://github.com/google-gemini/cookbook/tree/main/gemini-2/), in particular the [video understanding](../quickstarts/Video_understanding.ipynb) and the [spatial understanding](../quickstarts/Spatial_understanding.ipynb) ones.\n"
```
markmcd, line 667: This one doesn't link to a file so I think it's OK to leave as a full absolute URL.

e.g. if we use `/google-gemini/cookbook/...` then it breaks on Colab (where it uses `colab.research.google.com/github/org/repo/...`).

markmcd (approved):

PR #542, opened 2025-03-09
Set up a TS folder and add an explicitly typed chat example 

myanvoos:
OK -- I've been working on this for a few days now. I hope this will be the first example in what will be a growing collection of TS implementations!

This PR sets up a dedicated TS example folder, scaffolding it with a basic `package.json`, `README.md`, and `tsconfig.json` structure so that future TS contributions can just go in here. It also includes one initial example of a medieval castle analysis chat tool that explicitly exposes the SDK types (future TS examples probably don't need to be this explicit) and makes use of Gemini's image detection capability. I'm using the available castle.png sitting in the assets folder, but of course can always change this to something more exciting.

The TS folder is currently located inside the `examples` folder. I'm open to suggestions if you think this should be moved somewhere else @Giom-V  and @markmcd. Also, any feedback on structure, writing, or best practices would be much appreciated.

Edit: I'll go through the example and rewrite it slightly to better adhere to https://developers.google.com/style. Marking this as a draft for now.

### DIFF
--- a/examples/typescript/.gitignore
+++ b/examples/typescript/.gitignore
@@ -1,0 +1,2 @@
+node_modules
+package-lock.json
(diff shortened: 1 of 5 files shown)


myanvoos:
After the unified JS/TS SDK went public this is now no longer up-to-date! Closing this down now.

Issue #551, opened 2025-03-11
Book illustrations notebook doesn't load on GitHub


markmcd:
### Description of the bug:

We should reduce the amount of inline content (eg moving images out to GCS, reducing the number of examples, resizing before previewing)

ROHANDWIVEDI2005:
Hi @markmcd I think **resizing** would be the best option here because this will keep all the examples. moving them to **GCS** would add extra complexity

Swini27:
Hi @markmcd What do you think about moving ````Book_illustrations.ipynb```` to git lfs? We can just host this one file on the large file storage, while it would load on GitHub, it will also allow us to preserve the entire existing inline content.

william-Dic:
Hi @markmcd, I prepared two solutions for this issue; both are working well now. 

![Image](https://github.com/user-attachments/assets/ec93fd91-15a8-4636-b608-38fce5d21295)

The reason the notebook wasn't rendering correctly on GitHub was due to the large size of the image outputs embedded within it.

Here are the two effective solutions:

1. Reduce the number of generated images by changing _max_character_images_ and _max_chapter_images_ variables down to 1 or 2. Saving images as JPEG locally instead of PNG drastically reduces the image file size while keeping good visual quality, making the notebook render smoothly on GitHub.

rubiagatra:
This issue should be closed

PR #557, opened 2025-03-12
Fixed Book illustrations notebook doesn't load on GitHub Issue #551

william-Dic:
I've identified the cause of the rendering issue on GitHub—it was due to the large file sizes of image outputs embedded in the notebook.

I've tested two effective solutions, and both now work smoothly:

Image

1. **Reduce Image Quantity:** Lower the values for `_max_character_images_` and `_max_chapter_images_` to 1 or 2, minimizing embedded image data.

2. **Optimize Image Format:** Resize the images generated and saving images as JPEG with slightly lower quality locally instead of PNG makes the .ipynb size almost 10 times smaller than original one.


Giom-V:
Sorry, this was my mistake (but the notebook is still working using colab). And thanks for the fix but I had other plans to fix it (basically storing the output somewhere else as I just did on the image-out notebook) so I might not be able to take your fixes.

That said, I saw another example you want to add with your PR, so please keep that part and let's rename that PR.

william-Dic:
Thanks for reviewing this, @Giom-V!

The other example included in the PR relates to the fourth idea proposed for GSoC 2025—I found an efficient way to cache text from a 2-hour lecture, which turned out to be quite interesting. It's great to hear you've already planned an alternative solution for the notebook issue. If there's anything else I can assist with or contribute, please let me know. I'm looking forward to contributing more to the project!

Appreciate your feedback, and wish you all the best with your work!

Giom-V:
@william-Dic I'm closing the PR as I've already fixed the book illustration in #562. For the record I'm planning to rework the `.show()` function to be able to resize the images and reduce the size of the notebook while keeping the outputs.

Feel free to send your other ideas through a new PR. I checked it very quickly but it seems like an interesting use-case (but it will have to use the Gemini API instead of the Vertex ones to be here).

Issue #569, opened 2025-03-15
Enhance Error Handling and Secure API Key in callGemini Function


amanmogal:
### Description of the bug:

The function callGemini in the utils.gs file does not handle potential errors from network requests effectively, and there is a risk of API key exposure. Here are the main concerns:

1. Error Handling: The function does not handle possible errors like network failures or invalid responses from the API.
Steps to Reproduce the Issue:

1. Run the callGemini function with an invalid API endpoint or network conditions that cause a failure.
2. Observe that there is no error handling mechanism to catch and address the failure.

### Actual vs expected behavior:

- Error Handling:

- Actual: No error handling for network failures or invalid API responses.
- Expected: Proper error handling to log errors and return meaningful messages.

Bhavesh2k4:
I'd like to work on this issue. My plan is to:
1. Implement try/catch blocks for all API calls
2. Move the API key to auth headers instead of URL parameters
3. Add input validation for all functions
4. Add proper error logging

Let me know if you'd like me to proceed with these changes.

SaurabMishra12:
I've been working on this issue and here are some ideas I've come up with:

for Error Handling, you can wrap network requests in a try/catch block so that if something goes wrong, the script can catch the error and handle it gracefully instead of crashing. 
You can also set muteHttpExceptions: true to inspect response codes and decide how to proceed when you don't get a 200 status.
for API Key Security= Rather than hardcoding my API key in the code, it's better to move it to a secure storage solution. I'm using the Properties Service for now, but I recommend considering Google Cloud Secret Manager for even tighter security. This helps prevent accidental exposure of the key.
 Also, you can implement a retry mechanism with delays to handle intermittent failures. Plus, adding logging or notifications will alert you if repeated attempts fail, so you can jump in and troubleshoot before it becomes a bigger problem.

Let me know if you need updated codes for this.

PR #571, opened 2025-03-15
Enhance Error Handling in callGemini Function

amanmogal:
This PR addresses the issue #569 where the callGemini function in utils.gs does not handle potential errors from network requests effectively. The main concerns addressed in this PR are:

Error Handling:

1. The function did not previously handle possible errors like network failures or invalid responses from the API.
2. Added a try-catch block to handle potential errors, log the error details, and return a meaningful error message without causing the script to fail.

Changes Made:

Added error handling to the callGemini function to manage network failures and invalid API responses.

Steps to Reproduce the Issue:

1. Run the callGemini function with an invalid API endpoint or under network conditions that cause a failure.
2. Observe that the function now handles errors gracefully, logging the error and returning an appropriate message

Expected Behavior:

- The callGemini function should now catch and handle errors, logging details and returning user-friendly messages.

By merging this PR, the callGemini function will become more robust and maintainable, with improved error handling and secure API key management. #569

### DIFF
--- a/examples/Apps_script_and_Workspace_codelab/utils.gs
+++ b/examples/Apps_script_and_Workspace_codelab/utils.gs
@@ -95,6 +95,11 @@ function callGemini(prompt, temperature=0) {
     "contents": [
       {
         "parts": [
+          {function callGemini(prompt, temperature=0) {
+  const payload = {
+    "contents": [
+      {
+        "parts": [
           {
             "text": prompt
           },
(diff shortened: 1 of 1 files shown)


examples/Apps_script_and_Workspace_codelab/utils.gs
```diff
@@ -91,6 +91,11 @@ const WORKSPACE_TOOLS = {
 };
 
 function callGemini(prompt, temperature=0) {
+  const payload = {
```
Giom-V, line 94: Wihtout even running it, the code doesn't seem to be legit as you are not closing all your brackets and defining `payload` twice.

amanmogal: I mistakenly committed the code. Found error in its codelab file as well testing the new changes and will update the file

PR #574, opened 2025-03-17
poem generation and qa bot using gemini function calling  \

Pavansomisetty21:


Giom-V:
Hello @Pavansomisetty21,

I think it's a great example (which could benefit from speech generation when it will be available), but can you reformat it following the [template](https://github.com/google-gemini/cookbook/blob/main/quickstarts/Template.ipynb) and in particular using the API key stored in the colab secret?

I think the UX on colab could also be improved by using [colab forms](https://colab.sandbox.google.com/notebooks/forms.ipynb) like `MODEL_ID="gemini-2.0-flash" # @param ["gemini-2.0-flash-lite","gemini-2.0-flash","gemini-2.0-pro-exp-02-05"] {"allow-input":true}` instead of `input`. Maybe it would be worth having both versions (colab first with explanations, then a raw python one if someone wants to run it on their machine).

Pavansomisetty21:
Ok I will do it

manydevelopers1:
great work @Pavansomisetty21  and @Giom-V you must accept it ,was good work

kkorpal:
Hi @Pavansomisetty21  Thanks for this awesome contribution! 
To get this ready for merging, could you make a few quick updates to match our current repository standards?
SDK & Model: Please migrate to the new google-genai SDK and set the default model to gemini latest models.
Please work on @Giom-V comment (https://github.com/google-gemini/cookbook/pull/574#issuecomment-2765712383)

API Key: For security, use Colab secrets (userdata.get("GOOGLE_API_KEY")) instead of a hardcoded string.
Let us know if you need a hand with any of these updates!

Issue #584, opened 2025-03-19
"Browser as a tool" colab is going into error: 1 validation error for LiveConnectConfig


apelliciari:
### Description of the bug:

This example is giving the following error when run:

```
<ipython-input-7-cbc631909dde>:2: ExperimentalWarning: The live API is experimental and may change in future versions.
  async with client.aio.live.connect(model=MODEL, config=config) as stream:
---------------------------------------------------------------------------
ValidationError                           Traceback (most recent call last)
[<ipython-input-7-cbc631909dde>](https://localhost:8080/#) in <cell line: 1>()
      5     await stream_response(stream)
      6 
----> 7 await run()

3 frames
[/usr/local/lib/python3.11/dist-packages/pydantic/main.py](https://localhost:8080/#) in __init__(self, **data)
    212         # `__tracebackhide__` tells pytest and some other tools to omit this function from tracebacks
    213         __tracebackhide__ = True
--> 214         validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
    215         if self is not validated_self:
    216             warnings.warn(

ValidationError: 1 validation error for LiveConnectConfig
generation_config.response_modalities
  Extra inputs are not permitted [type=extra_forbidden, input_value=['TEXT'], input_type=list]
    For further information visit https://errors.pydantic.dev/2.10/v/extra_forbidden
```

william-Dic:
Hi @apelliciari,

It looks like the issue is due to an update in how the configuration is passed.

If you update your `config` to the following format, it should resolve the problem you're experiencing:

```python
config = {
    'response_modalities': ['TEXT'],
    'tools': [
        {'google_search': {}},
    ],
}
```

Here's the updated output after applying this change:

![Image](https://github.com/user-attachments/assets/33ced1c3-b05e-4c1f-bcef-8975bfe813ef)

apelliciari:
thanks, in the PR you should also update the config on the `load_page_def` cell

```
load_page_def = types.Tool(functionDeclarations=[
    types.FunctionDeclaration.from_callable(client=client, callable=load_page)]).model_dump(exclude_none=True)

config = {
    'generation_config' : {
        'response_modalities': ['TEXT']
    },
    'tools': [
        load_page_def,
    ],
    'system_instruction': """Your job is to answer the users query using the tools available.

First determine the address that will have the information and tell the user. Then immediately
invoke the tool. Then answer the user.
"""
}


async def run():
[...]

PR #585, opened 2025-03-19
Changed the config format and fixed issue #584

william-Dic:
That issue is due to an update in how the configuration is passed.

`config` can work with the following format:

```python
config = {
    'response_modalities': ['TEXT'],
    'tools': [
        {'google_search': {}},
    ],
}
```


Giom-V:
Thanks for the submission @william-Dic but I think @markmcd already fixed it in #598.

PR #596, opened 2025-03-23
Create python-package-conda.yml

C4w112a:
All operation python


Giom-V:
Why should we build a conda package?

C4w112a:
> Why should we build a conda package?

For encable tool transformation and for on untubuntu

markmcd:
We don't ship a Python package in this repository, so no need for any Conda packaging. Please comment again if you think we're mistaken but otherwise this repo is for sharing documentation.

PR #599, opened 2025-03-24
Marking links as external so they open in a new tab

Giom-V:


Giom-V:
`{:target="_blank"}` is another option. I'm not sure what the difference is...

Giom-V:
It's not working as expected.

PR #592, opened 2025-03-22
Updated quickstarts/Video.ipynb to use google-genai SDK

prajwxlk:
- Updated the code to use the latest sdk
- Implemented timeout according to the latest sdk

PR #616, opened 2025-03-30
migrated to latest sdk

andycandy:
- Updated module google-generativeai to use google-genai
- Refactored imports
- nbfmt
- lint checked


Giom-V:
Thanks @andycandy but we already have another PR for this file (#592). So since you already did a lot, I'll close and let @prajwxlk have some work as well :)

andycandy:
Oh apologies, I forgot to check

Issue #617, opened 2025-03-31
Refactor `Basic_Code_Generation.ipynb` to execute the generated code


Giom-V:
### Description of the feature request:

At the moment the code generated in the `Basic_Code_Generation.ipynb` notebook is not executed and instead copied manually in a code cell. It would be way bette to get the code from response.text and use exec() to run it.

Cf. https://github.com/google-gemini/cookbook/pull/515#discussion_r2009846237 for a related dicsussion

andycandy:
I solved it simply by changing the model, for now gemini-2.0-flash prints indentation issues, but the other models work fine

andycandy:
Solved this issue in PR#618

PR #618, opened 2025-03-31
Fixed issue#617

andycandy:
- Added regex extraction of code from the response
- Executed the extracted code and it worked well
- nbfmt and lint checked

### DIFF
--- a/examples/prompting/Basic_Code_Generation.ipynb
+++ b/examples/prompting/Basic_Code_Generation.ipynb
@@ -170,24 +170,19 @@
       "cell_type": "code",
       "execution_count": 12,
       "metadata": {
-        "id": "CHTdAVE0pIFf",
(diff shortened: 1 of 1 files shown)


Giom-V:
Thanks for the fix @andycandy! 
Just nitpicking as I really appreciate all you've done for the cookbook, but I entered #617 as a "good first issue" for someone who never worked on the Gemini API to work on. So maybe next time, if you see that label, can you wait a couple of days before submitting a fix?

Giom-V (approved):

andycandy:
You're right, @Giom-V , I do apologize for not paying closer attention to the labels, I'll make sure to double-check in the future and give those "good first issues" some space.

Giom-V:
> You're right, @Giom-V , I do apologize for not paying closer attention to the labels, I'll make sure to double-check in the future and give those "good first issues" some space. 

No worries. That was also my first one, so next time I'll advertise it better as well.

PR #619, opened 2025-03-31
Close links html tags on "open in colab" buttons

Giom-V:


Giom-V:
The 2 fails are "work as intended"

patrickloeber (approved):

PR #629, opened 2025-04-01
Add new example ( Google search with Gemini )

omkenge:
By enabling Google Search, Gemini can provide more accurate and timely responses.

### DIFF
--- a/examples/gemini_google_search.ipynb
+++ b/examples/gemini_google_search.ipynb
@@ -1,0 +1,211 @@
+{
+  "nbformat": 4,
+  "nbformat_minor": 0,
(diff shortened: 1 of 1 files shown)


Giom-V:
@omkenge Thanks for the submission. I think it would be better to add those examples in https://github.com/google-gemini/cookbook/blob/main/quickstarts/Search_Grounding.ipynb so that the different ways to use grounding (chat or unary) are showcased in the same notebook. Can you do that?

omkenge:
@Giom-V Sure I will do this

omkenge:
new PR raised sorry for confusion

PR #645, opened 2025-04-03
Updated Safety.ipynb to use google-genai

omkenge:
Updated module google-generativeai to use google-genai
Refactored imports
Modified how configs are being assigned to models


omkenge:
Already PR Exist #645

PR #655, opened 2025-04-04
Formatting Apollo11.ipynb Notebook

omkenge:


examples/Apollo_11.ipynb
```diff
@@ -1,5 +1,15 @@
 {
```
andycandy, line 1: <pre class="ql-syntax" spellcheck="false">Line #13.       "httpOptions": {"timeout": 18000}
</pre><p>I think the more appropriate form would be <code>10 * 60 * 1000</code>, that'd help developers understand the duration unit better</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google-gemini/cookbook/pull/655/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

andycandy:
This notebook was updated to google-genai few weeks before as well, please rename your PR's appropriately

examples/Apollo_11.ipynb
omkenge: <p>Done</p>

omkenge:
@andycandy Thanks for your suggestions :)

Giom-V:
Thanks for the submission @omkenge but I don't really see added value with updating the notebook that way. It might be a personal opinion but I usually prefer when all values are indicated in the generation call instead of having to refer to a previous variable. And I'm not a fan of variables that are only used once.

PR #672, opened 2025-04-05
Update `google-generativeai` to `google-genai` in `New_in_002.ipynb` 

nsbg:
@Giom-V 

You mentioned in [#446](https://github.com/google-gemini/cookbook/issues/446): 'No need to migrate, I'll merge it into an "advanced parameter notebook" (or make this one instead ;))', but I couldn't find the advanced parameter notebook yet, so I started working on this file instead. 

It's about halfway done, but I opened this PR to check whether I should continue or stop here. 

I'd appreciate your thoughts!

### DIFF
--- a/examples/prompting/Providing_base_cases.ipynb
+++ b/examples/prompting/Providing_base_cases.ipynb
@@ -11,9 +11,8 @@
     },
     {
       "cell_type": "code",
-      "execution_count": null,
+      "execution_count": 1,
       "metadata": {
-        "cellView": "form",
         "id": "fKqk8LWo_TCU"
       },
       "outputs": [],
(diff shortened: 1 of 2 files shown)


andycandy:
I think it makes more sense to create a new notebook that highlights the capabilities of the latest models, like Gemini 2.5. The current notebook was built back when Gemini 2.0 had just launched, so showcasing the improvements with a fresh example feels like a better direction.

andycandy:
You can always refer to ./quickstarts/template.ipynb for creating new notebooks

nsbg:
@andycandy Thank you for you suggestion 😊 I should try working on it in the direction you suggested.

Giom-V:
@nsbg Thanks a lot for the contribution!

As @andycandy said, I think it would be better to create a new notebook as I had planned. That said, it's going to be a landfield as there's a lot to explain (and a lot to come), and a need to explain it right, which means I might ask you for a lot of rework and we're going to have to go through a lot of reviews (I might even ask my TW colleagues to have a look).

So I would not forbid to work on that, on the contrary, but be warned that it's a tough one so be warned...

Issue #681, opened 2025-04-07
Cleanup: Please move the websocket examples from the deprecated gemini-2 directory into quickstarts with the opthers.


MarkDaoust:
### Description of the feature request:

See title.

Giom-V:
Fixed by #687

MarkDaoust:
Thanks!

PR #685, opened 2025-04-07
Move WebSocket examples to `quickstarts/`

KSruthiVel:
This PR moves all WebSocket-related examples from the deprecated `gemini-2/websockets/` directory to `quickstarts/websockets/` to align with the current project structure.

This PR addresses #681


Giom-V:
Thanks for the submission @KSruthiVel but it's slightly more complicated than just moving files as there are links to fix, placeholders to create so that old links to the cookbook won't be broken. You can check #687 for reference on how it is done as I was already working on that.

PR #691, opened 2025-04-09
Update import advice in template

markmcd:
Main points here are fixing the import guidance and explicitly calling out that a guide needs text (seems to be a common issue, notebooks with nothing but code).


quickstarts/Template.ipynb
```diff
@@ -75,10 +75,10 @@
       },
```
Giom-V, line 75: <pre class="ql-syntax" spellcheck="false">Line #1.    %pip install -U -q "google-genai~=1.0"  # Install the Python SDK
</pre><p>Any reason why you prefer it like this? I've updated the whole cookbook to use <code>&gt;=1.0.0"</code> recently... Maybe we should discuss it? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google-gemini/cookbook/pull/691/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

Giom-V (approved):

PR #643, opened 2025-04-03
Updated Providing_base_cases.ipynb to use google-genai

Shi-pra-19:
Updated module google-generativeai to use google-genai
Refactored imports

PR #705, opened 2025-04-13
Add `Add_prefixes.ipynb` in examples/prompting directory

nsbg:
I checked the Gemini API docs, and it seems like the 'Add prefixes' section under [prompting strategies](https://ai.google.dev/gemini-api/docs/prompting-strategies#add-prefixes) is the only part missing an example file. So, I've added a simple example file for prefixes. 

I carefully reviewed the PR and repository this time, and I hope this file is helpful for the cookbook repository 😇 Any feedback is welcome!


andycandy:
Thank you for your contribution but `base cases` and `New in 002` notebooks have already been addressed in PRs #643 n #672.

nsbg:
The newly added `Add_prefixes.ipynb` file is unrelated to the two issues but the commit history seems to be connected. What should I do if I want to add new example file only?

andycandy:
I believe the best thing to do would be to create a new PR, but if you would like to continue this pr, here is how you can do it:
- Reset the two files:
`git checkout origin/main --examples/prompting/Providing_base_cases.ipynb` same for 002
- Stage the changes then commit

PR #716, opened 2025-04-15
add Flash 2.0 to caching notebook

nikitamaia:
- Add gemini-2.0-flash-001 to list of supported models for context caching
- Set gemini-2.0-flash-001 as default model for caching notebook
- Set gemini-2.0-flash-001 as default model for get started notebook

### DIFF
--- a/quickstarts/Caching.ipynb
+++ b/quickstarts/Caching.ipynb
@@ -66,16 +66,7 @@
       "metadata": {
         "id": "5yRLvyPhrXSf"
       },
-      "outputs": [
-        {
-          "name": "stdout",
-          "output_type": "stream",
-          "text": [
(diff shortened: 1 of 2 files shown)


Giom-V (approved):

quickstarts/Caching.ipynb
```diff
@@ -66,16 +66,7 @@
       "metadata": {
```
Giom-V, line 66: <p>Nit: the number of token seems to be different from what it was with 1.5. You should maybe update the text.</p><p> </p><p>You should also remove the "75%" as it's not true anymore and just say something like "way cheaper" so that it always stays true.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google-gemini/cookbook/pull/716/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='25'/>

nikitamaia: <p>Updated the numbers and added an extra sentence to indicate which metadata fields to look at</p>

Giom-V (dismissed):
Approved but I think there's something fishy going on so I'll double-check tomorrow.

lucianommartins (approved):
LGTM, approved.

PR #767, opened 2025-05-15
Text update: remove reference and link to unavailable content

jurgenizer:
Remove reference and link to AI Studio's live demo as it is no longer available.


Giom-V:
Thanks for catching the problem @jurgenizer, I wasn't aware the app had been taken down. I checked with the team and they are going to release a new version, so I told them to make sure the current url would point to that new version. I'll close the PR when it will be the case (keeping it open so that I don't forget to double-check)

Giom-V:
Closing as this is now fixed.

PR #637, opened 2025-04-02
Update package to google-genai in examples/Search_reranking_using_embeddings.ipynb

Jenniebn:
Update package to google-genai in `examples/Search_reranking_using_embeddings.ipynb`

- Updated google-generativeai to use google-genai package instead.
- Updated outputs and tested locally.
- Refactored imports
- Restructured the sections and updated instructions
- Formatted with nbfmt and nblint tool

PR #768, opened 2025-05-18
Update Search_reranking_using_embeddings notebook to use google-genai SDK (#446)

### LINKED ISSUE #446: Update everything to use google-genai SDK

sbeeredd04:
Updated the Search_reranking_using_embeddings.ipynb notebook to use the new google-genai SDK v1.0+. Changes include:

- Updated imports from `google.generativeai` to `google import genai`
- Migrated function calling implementation to use the new API pattern
- Updated configuration syntax for chat creation and tools configuration
- Simplified automatic function calling implementation to use the new approach
- Changed embedding API calls to use the new format
- Updated explanatory text to reflect the new API patterns and methods
- Fixed manual function calling implementation to use the updated Part.from_function_response approach

All functionality remains the same, but the code now follows the recommended patterns from the migration guide.

Fixes #446


andycandy:
PR #637 is already actively being worked on to update this notebook.

PR #776, opened 2025-05-22
Adding a python script for Lyria RealTime

Giom-V:


markmcd (dismissed):
FWIW, the audio plays over itself when you press "Run all", and the "wait 5 seconds" thing didn't work when I tried it.

markmcd (approved):

Giom-V:
> FWIW, the audio plays over itself when you press "Run all", and the "wait 5 seconds" thing didn't work when I tried it. 

Sadly it doesn't work all the time :(

PR #788, opened 2025-05-29
feat: Enable OSC control for Lyria real-time script

didier-brown:
I've modified `quickstarts/Get_started_LyriaRealTime.py` to replace command-line input with an OSC (Open Sound Control) server for controlling Lyria music generation.

Key changes:
- I integrated the `python-osc` library to create an OSC server.
- The OSC server runs in a separate thread and communicates with the main asyncio loop via an `asyncio.Queue`.
- I implemented handlers for the following OSC messages:
    - `/lyria/setPrompts` (string "Prompt1:Weight1,..."): Sets weighted musical prompts.
    - `/lyria/play`: Starts/resumes playback.
    - `/lyria/pause`: Pauses playback.
    - `/lyria/stop`: Stops generation and exits.
    - `/lyria/bpm` (int/float or "AUTO"): Sets BPM.
    - `/lyria/scale` (string or "AUTO"): Sets musical scale.
- Original audio output via PyAudio is maintained.
- I added `quickstarts/osc_test_client.py` for testing OSC functionality.
- I updated dependencies to include `python-osc`.

This modification allows for external control of the Lyria script, enabling integration with applications like TouchDesigner as outlined in the project plan.

### DIFF
--- a/quickstarts/Get_started_LyriaRealTime.py
+++ b/quickstarts/Get_started_LyriaRealTime.py
@@ -19,7 +19,7 @@
 To install the dependencies for this script, run:
 
 ```
-pip install pyaudio websockets
+pip install pyaudio websockets python-osc
 ```
 
 Before running this script, ensure the `GOOGLE_API_KEY` environment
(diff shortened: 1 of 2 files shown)


Giom-V:
Hello @didier-brown and thanks a lot for the submission.

While as an individual I greatly appreciate your submission since it opens a lot of home automation possibilities, as the cookbook gatekeeper I don't think we should update the code sample. The reason for that is because this code sample needs to be as simple as possible so that anybody can understand it and use it for their own use-cases, the same way you just did. Honestly, I already think I over-complexified it...

Maybe we should make it an example instead? What about we move that into a folder under examples/ along with a README that explains how it works?

PR #802, opened 2025-06-14
I've added a new JavaScript function called `parseQueryString` to `ex…

Mwessc:
…amples/Apps_script_and_Workspace_codelab/parseQueryString.js`.

This function takes a URL query string and turns it into an object. The keys in this object are the query parameters, and their values are the corresponding values from the query string.

Here's what `parseQueryString` can do:
- It works whether or not the query string starts with a '?'.
- It correctly handles both single and multiple parameters.
- If you have the same parameter multiple times, it will store all its values in an array.
- It can deal with URL-encoded characters in both parameter names and values.
- If you give it an empty, null, or undefined query string, it will return an empty object.
- It can handle parameters that don't have values (like `key=`) and parameters without an equals sign (like `key`, which will result in `key: ''`).

I've also included unit tests for this function in `examples/Apps_script_and_Workspace_codelab/parseQueryString.test.js`. These tests cover a wide variety of situations, including:
- Empty and null/undefined inputs.
- Simple and multiple parameters.
- Parameters with multiple values.
- URL-encoded characters.
- Query strings with and without a leading '?'.
- Parameters with no values or no equals sign.
- More complex combinations of these scenarios.

You can run these tests in a basic JavaScript environment, like a browser console or Node.js, and they will log their results.

### DIFF
--- a/examples/Apps_script_and_Workspace_codelab/parseQueryString.js
+++ b/examples/Apps_script_and_Workspace_codelab/parseQueryString.js
@@ -1,0 +1,44 @@
+/**
+ * Parses a query string into an object.
+ *
+ * @param {string} queryString The query string to parse.
+ * @return {object} An object containing the parsed key-value pairs.
+ */
+function parseQueryString(queryString) {
+  const params = {};
+  if (!queryString) {
+    return params;
+  }
+
+  // Remove leading '?' if present
+  if (queryString.startsWith('?')) {
+    queryString = queryString.substring(1);
+  }
+
+  if (!queryString) { // Check again in case query string was only "?"
+    return params;
+  }
+
+  const pairs = queryString.split('&');
+
+  for (let i = 0; i < pairs.length; i++) {
+    const pair = pairs[i].split('=');
+    const key = decodeURIComponent(pair[0]);
+    let value = '';
+    if (pair.length > 1) {
+      value = decodeURIComponent(pair[1] || '');
+    }
+
+    if (key in params) {
+      if (Array.isArray(params[key])) {
+        params[key].push(value);
+      } else {
+        params[key] = [params[key], value];
+      }
+    } else {
+      params[key] = value;
+    }
+  }
+
+  return params;
+}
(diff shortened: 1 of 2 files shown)


andycandy:
I'm a bit unsure about how this addition fits into the cookbook repo and what purpose it serves. Could you clarify that a bit?

Giom-V:
@Mwessc While I can see that the function can be useful in some case, I fail to see what it brings to the appscript integration of Gemini. Can you clarify?

Mwessc:
The parseQueryString function brings significant utility to the App Script integration of Gemini by simplifying how your scripts can interact with web-based APIs and data, which is often the case when dealing with services like Gemini. Here's a breakdown of its value:

Simplified Parameter Handling for Gemini Requests: When interacting with Gemini through web service calls (which is common in App Script), you often need to construct URLs with specific parameters to define your query. This function allows you to easily take a query string, potentially built dynamically within your App Script, and convert it into a structured JavaScript object. This makes it much easier to access and manage the individual parameters you need to send to Gemini.

Parsing Gemini API Response Data: While the direct response from a Gemini API might be in JSON or another structured format, sometimes APIs (including those Gemini might interact with internally or that your App Script might use in conjunction with Gemini) pass data back in the form of URLs with query parameters. This function provides a straightforward way to extract information encoded in this manner.

Flexibility in Handling User Input: If your App Script allows users to provide URLs related to their interaction with Gemini, this function can robustly parse those URLs, regardless of whether they start with a ?, have multiple parameters, or contain URL-encoded characters. This ensures your script can reliably extract the relevant information from user-provided input.

Robustness and Error Handling: The function is designed to handle various edge cases like empty strings, null or undefined input, and parameters without values. This makes your App Script more resilient to unexpected or poorly formatted data, leading to a more stable integration with Gemini.

In essence, parseQueryString acts as a handy utility to bridge the gap between URL-based data (common in web interactions) and the structured data formats that are easier to work with within your App Script logic when integrating with Gemini. It reduces the need for manual string manipulation and makes your code cleaner, more readable, and less prone to errors when dealing with URL parameters.

PR #804, opened 2025-06-16
Delete quickstarts/Tuning.ipynb

MarkDaoust:


Giom-V:
We could also replace it by an empty notebook that says to use Vertex.

MarkDaoust:
> Empty notebook

Good idea, we can point it to here:

 https://cloud.google.com/vertex-ai/generative-ai/docs/models/gemini-use-supervised-tuning#google-gen-ai-sdk

PR #810, opened 2025-06-23
Add a quick multi-file example

MarkDaoust:
Note: Using async here doesn't speed it up at all.


Giom-V:
@MarkDaoust Great addition, but can you run the formatting script?

MarkDaoust:
Done.

PR #811, opened 2025-06-24
Use known/tested models for browser demos

markmcd:
These are known to be stable, so using as defaults for I/O workshops.


irbg (dismissed):

markmcd:
Lint is failing with false positive. 'native' is a model name so we can't change it.

irbg (approved):

PR #824, opened 2025-07-02
Modifications to led controller example

SpiderDerp:
On newer module versions, MISO and MOSI pins are referred to as DO and DI.

HTTP requests require a content-length header to prevent a 411 error.

The code was originally made to record audio when the button is not pressed, but it was changed to do the opposite.

### DIFF
--- a/examples/iot/esp32/voice_led_controller/led_controller.ino
+++ b/examples/iot/esp32/voice_led_controller/led_controller.ino
@@ -305,7 +305,8 @@ void sendAudio() {
     file.close();
     SD.end(); // Close the SD connection after reading the file
     
-    int httpCode = http.POST(jsonString);
+    http.addHeader("Content-Length", String(strlen(jsonString))); //http requests require Content-Length header to prevent http 411 error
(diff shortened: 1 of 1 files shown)


SpiderDerp:
I signed the CLA after being notified.

examples/iot/esp32/voice_led_controller/led_controller.ino
```diff
@@ -418,7 +419,7 @@ void setup() {
 }
 
 void loop() {
-  if (digitalRead(BUTTON_PIN) == LOW) {
+  if (digitalRead(BUTTON_PIN) == HIGH) { //check if button is pressed
```
PaulTR, line 422: In the diagram there's a pull down resistor with the button - guessing you're using a different setup? This will likely change for people depending on what button they're using.

SpiderDerp: I followed the diagram with the pull-down resistor and the button, but the recording that is supposed to occur upon button press only worked for me when I modified this line. Otherwise, it would record when the button is not being pressed.
It could be possible that I made a wiring mistake, so feel free to ignore this change.

PR #833, opened 2025-07-04
Updating Get_Started file

andycandy:


quickstarts-js/Get_Started.js
```diff
@@ -1,158 +1,1291 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* Markdown (render)
+# Gemini API: Getting started with Gemini models
+
+The new **[Google Gen AI SDK](https://googleapis.github.io/js-genai)** provides a unified interface to [Gemini models](https://ai.google.dev/gemini-api/docs/models) through both the [Gemini Developer API](https://ai.google.dev/gemini-api/docs) and the Gemini API on [Vertex AI](https://cloud.google.com/vertex-ai/generative-ai/docs/overview). With a few exceptions, code that runs on one platform will run on both. This notebook uses the Developer API.
```
Giom-V: Let's drop the "new". It's not new anymore :)

quickstarts-js/Get_Started.js
```diff
@@ -1,158 +1,1291 @@
...
+* Counting tokens
+* Setting system instructions
+* Configuring safety filters
+* Initiating a multi-turn chat
+* Controlling generated output
+* Using function calling
+* Generating a content stream
+* Using file uploads
+
+More details about this new SDK on the [documentation](https://ai.google.dev/gemini-api/docs/sdks).
```
Giom-V: Same comment about "new"

quickstarts-js/Get_Started.js
```diff
@@ -1,158 +1,1291 @@
...
+
+## Setup
+### Install SDK and set-up the client
+*/
+
```
Giom-V, line 54: Add an explanation about saving the API key in the environmental variables and never save it in the front-end. Use https://ai.google.dev/gemini-api/docs/api-key#set-api-env-var as a reference (and add a link)

quickstarts-js/Get_Started.js
```diff
@@ -1,158 +1,1291 @@
...
+/* Markdown (render)
+## Count tokens
+
+Tokens serve as the fundamental input units for Gemini models. You can use the `countTokens` method to calculate the number of input tokens prior to making a request to the Gemini API, and the `totalTokens` property to access the total token count after the request is processed.
+*/
+
+// [CODE STARTS]
+const resp2 = await ai.models.countTokens({
```
Giom-V: Try to stay consistent with the response variables. Either use `response` or `resp` all the time, or increment them all, but stay consistent.

quickstarts-js/Get_Started.js
```diff
@@ -1,158 +1,1291 @@
...
+/* Output Sample
+
+8
+
+*/
+
+/* Markdown (render)
+## Send multimodal prompts
+
+Use Gemini 2.0 model (`gemini-2.0-flash`) or a later version, a multimodal model that supports multimodal prompts. You can include text, PDF documents, images, audio and video in your prompt requests and get text or code responses.
```
Giom-V: Rephrase that, maybe just say "Gemini models are all multimodal models"

quickstarts-js/Get_Started.js
```diff
@@ -1,158 +1,1291 @@
...
+* Using file uploads
+
+More details about this new SDK on the [documentation](https://ai.google.dev/gemini-api/docs/sdks).
+
+## Setup
+### Install SDK and set-up the client
+*/
+
+// [CODE STARTS]
+module = await import("https://esm.sh/@google/genai@1.4.0");
```
Giom-V, line 56: ERROR: TypeError: Failed to fetch dynamically imported module: https://esm.sh/@google/genai@1.4.0

Giom-V: I think we have a problem here. I did not realize it earlier but this way of importing is kinda weird. We should use `import {GoogleGenAI} from '@google/genai';`, except it doesn't work. Let's think of a workaround, because the code in the sample/notebook should be as close as possible to the documentation.

quickstarts-js/Get_Started.js
```diff
@@ -1,158 +1,1291 @@
...
+
+
+/* Markdown (render)
+In this second example, you'll open a previously saved image, create a thumbnail of it and then generate a short blog post based on the thumbnail, displaying both the thumbnail and the generated blog post.
+*/
+
+
+// [CODE STARTS]
+const response = await ai.models.generateContent({
+  model: "gemini-2.0-flash",
```
Giom-V: Use MODEL_ID

quickstarts-js/Get_Started.js
```diff
@@ -1,158 +1,1291 @@
...
+/* Markdown (render)
+## Send multimodal prompts
+
+Use Gemini 2.0 model (`gemini-2.0-flash`) or a later version, a multimodal model that supports multimodal prompts. You can include text, PDF documents, images, audio and video in your prompt requests and get text or code responses.
+
+In this first example, you'll download an image from a specified URL, save it as a byte stream and then write those bytes to a local file named `jetpack.png`.
+*/
+
+// [CODE STARTS]
+const IMAGE_URL = "https://cors-anywhere.herokuapp.com/https://storage.googleapis.com/generativeai-downloads/data/jetpack.png";
```
Giom-V: We need to use the images stored in our cloud bucket directly: https://storage.googleapis.com/generativeai-downloads/data/jetpack.png
Worst case, let's add the images on github.

quickstarts-js/Get_Started.js
```diff
@@ -1,158 +1,1291 @@
...
-
-/* Output Sample
-
-I will generate a 3D rendering of a whimsical scene: a pink pig with small, feathery wings gracefully soaring above a vibrant, futuristic cityscape. The city below will feature sleek, silver buildings intertwined with lush green vegetation and softly glowing lights. The pig will be wearing a dapper black top hat perched jauntily on its head.
-
-
-<img src="https://i.ibb.co/4DMSgyf/download.png"/>
-
-*/
+/*
```
Giom-V, line 1: General comment: Does the applet work all the time for you? For me I often have to refresh it multiple times before it displays its content, which is not good. I'm under the impression that it might be because it takes too much time retrieving the content from github. Can you investigate?

quickstarts-js/Get_Started.js
```diff
@@ -1,158 +1,1291 @@
...
+2.  &quot;You know, for an infinite expanse of spacetime, you&#x27;re surprisingly petty.&quot;
+
+*/
+
+/* Markdown (render)
+## Start a multi-turn chat
+
+The Gemini API enables you to have freeform conversations across multiple turns.
+
+Next you'll set up a helpful coding assistant:
```
Giom-V: "Next" feels weird. Maybe "Let's set-up..."

Giom-V: Also, I feel like the example is way too long, maybe come with something more simple (like just "talk like a pirate" or "You are an English teacher, tell me about the mistakes I make when talking to you" ).

quickstarts-js/Get_Started.js
```diff
@@ -1,158 +1,1291 @@
...
+        self.assertTrue(is_leap_year_concise(2024))
+        self.assertFalse(is_leap_year_concise(2023))
+
+
+# This allows running the tests directly from the script
+if __name__ == &#x27;__main__&#x27;:
+    unittest.main(argv=[&#x27;first-arg-is-ignored&#x27;], exit=False) # Use exit=False to allow running in interactive environments
+```
+
+### How to Run the Tests:
```
Giom-V: I think we need to change the css so that the titles in the ouputs are displayed in a different was than the ones from the notebook (maybe without colors and smaller).

quickstarts-js/Get_Started.js
```diff
@@ -1,158 +1,1291 @@
...
+
+#### 1. Save the chat history
+*/
+
+// [CODE STARTS]
+chatHistory = chat.getHistory()
+// [CODE ENDS]
+
+/* Markdown (render)
+#### 2. Resume later
```
Giom-V, line 325: Why is this mixing a code example and a cooking one?

quickstarts-js/Get_Started.js
```diff
@@ -1,158 +1,1291 @@
...
+    console.image(part.inlineData.data, "image/png");
+  }
+}
+// [CODE ENDS]
+
+/* Output Sample
+
+I will generate a 3D rendering of a whimsical scene. The central figure will be a pink pig, complete with small, delicate wings and a dapper grey top hat perched jauntily on its head. This unusual creature will be soaring above a sprawling futuristic cityscape. The city will feature sleek, modern buildings with sharp angles and glowing accents, but it will also be integrated with lush greenery, with trees and vines growing on the structures, creating a unique blend of nature and technology. The overall color palette will be vibrant, with the pink of the pig contrasting against the metallic and green hues of the city below.
+
+<img src="https://iili.io/FcTOzfj.png" alt="FcTOzfj.md.png" border="0">
```
Giom-V: Use https://storage.googleapis.com/generativeai-downloads/images/flying_pig.png

quickstarts-js/Get_Started.js
```diff
@@ -1,158 +1,1291 @@
...
+*/
+
+// [CODE STARTS]
+const response = await ai.models.generateContentStream({
+    model: "gemini-2.5-flash",
+    contents: "Tell me a story about a lonely robot who finds friendship in a most unexpected place.",
+});
+
+for await (const chunk of response) {
+    console.log(chunk.text);
```
Giom-V: I think you should add something (like "---") between the chunks so we can differentiate them

quickstarts-js/Get_Started.js
```diff
@@ -1,158 +1,1291 @@
...
+*/
+
+/* Markdown (render)
+### Upload text file
+
+Let's start by uploading a text file. In this case, you'll use a 400 page transcript from [Apollo 11](https://www.nasa.gov/history/alsj/a11/a11trans.html).
+*/
+
+// [CODE STARTS]
+TEXT_URL = "https://gist.githubusercontent.com/andycandy/1bd80850b4b80aa0608227e672015ad3/raw/b2a8008873dcd3dc6bf5740c1c05b1063f4952b0/apollo.txt"
```
Giom-V: We need to use https://storage.googleapis.com/generativeai-downloads/data/a11.txt

quickstarts-js/Get_Started.js
```diff
@@ -1,158 +1,1291 @@
...
+uploadResult = await ai.files.upload({
+  file: blob,
+  mimeType,
+});
+
+response = await ai.models.generateContent({
+  model: MODEL_ID,
+  contents: [
+    { fileData: { fileUri: uploadResult.uri, mimeType, } },
+    { text: "\n\nCan you give me a summary of this information please?" }
```
Giom-V: Can you try to prompt it for a shorter output?

quickstarts-js/Get_Started.js
```diff
@@ -1,158 +1,1291 @@
...
+/* Markdown (render)
+### Upload a PDF file
+
+This PDF page is an article titled [Smoothly editing material properties of objects](https://research.google/blog/smoothly-editing-material-properties-of-objects-with-text-to-image-models-and-synthetic-data/) with text-to-image models and synthetic data available on the Google Research Blog.
+
+Firstly you'll download a the PDF file from an URL and save it locally as "article.pdf
+*/
+
+// [CODE STARTS]
+pdfUrl = "https://cors-anywhere.herokuapp.com/https://storage.googleapis.com/generativeai-downloads/data/Smoothly%20editing%20material%20properties%20of%20objects%20with%20text-to-image%20models%20and%20synthetic%20data.pdf";
```
Giom-V: Needs to be https://storage.googleapis.com/generativeai-downloads/data/Smoothly%20editing%20material%20properties%20of%20objects%20with%20text-to-image%20models%20and%20synthetic%20data.pdf

quickstarts-js/Get_Started.js
```diff
@@ -1,158 +1,1291 @@
...
+*   **Applications:**
+    *   Facilitates design mock-ups (e.g., visualizing room repainting, new product designs).
+    *   Enables 3D consistent material edits by integrating with Neural Radiance Fields (NeRF) for synthesizing new views of an edited scene.
+
+*/
+
+/* Markdown (render)
+### Upload an audio file
+
+In this case, you'll use a [sound recording](https://www.jfklibrary.org/asset-viewer/archives/jfkwha-006) of President John F. KennedyÃ¢â‚¬â„¢s 1961 State of the Union address.
```
Giom-V, line 863: Encoding problem after "Kennedy". And we must not use a proxy.

quickstarts-js/Get_Started.js
```diff
@@ -1,158 +1,1291 @@
...
-/* Output Sample
-
-I will generate a 3D rendering of a whimsical scene: a pink pig with small, feathery wings gracefully soaring above a vibrant, futuristic cityscape. The city below will feature sleek, silver buildings intertwined with lush green vegetation and softly glowing lights. The pig will be wearing a dapper black top hat perched jauntily on its head.
-
-
-<img src="https://i.ibb.co/4DMSgyf/download.png"/>
-
-*/
+/*
+ * Copyright 2025 Google LLC
```
Giom-V, line 2: Other general comment. The text in the markdown is slightly too small. I checked the css and it uses the same format as the outputs. They should be separated and the explanations should be slightly bigger than the outputs.

quickstarts-js/Get_Started.js
```diff
@@ -1,158 +1,1291 @@
...
+*   **Blending/Pureeing:** This is a major distinction. The Allrecipes recipe explicitly instructs the user to purÃ©e the soup until smooth using a countertop or immersion blender. The Food.com recipe does not mention blending, suggesting a chunkier soup texture.
+*   **Roux Integration:** In the Food.com recipe, the flour-butter roux is prepared separately and then whisked into the boiling broth and vegetables. The Allrecipes recipe also prepares a roux (or bÃ©chamel with milk) separately but adds it to the already pureed soup base.
+*   **Order of Operations:** The Food.com recipe adds the half-and-half at the very end after the soup has thickened. The Allrecipes recipe adds the thickened milk mixture (roux with milk) to the soup base, then seasons it.
+
+In summary, the Allrecipes soup is designed to be a smooth, pureed soup with a stronger broccoli flavor due to the higher broccoli-to-broth ratio and includes celery for additional aromatic depth, using milk for its creaminess. The Food.com recipe appears to yield a soup with a more rustic, possibly chunkier texture, relying on half-and-half for richness and a larger volume of broth.
+
+*/
```
Giom-V, line 1098: Add a "What's next" section at the end to conclude.

quickstarts-js/Get_Started.js
andycandy: Yes, it consistently works for me on the first attempt. Considering that you're also encountering the "dynamic import" issue, the problem might be due to a slow internet connection or a related network issue.

quickstarts-js/Get_Started.js
andycandy: I agree, we should add "md" and "js" class names to the output cells based on their type. That way, it’ll be much easier to handle them consistently.

quickstarts-js/Get_Started.js
andycandy: Ah, it seems the markdown cell explaining the schema between these two wasn’t added correctly.

andycandy: I'll fix it

quickstarts-js/Get_Started.js
andycandy: Alright, I'll try to come up with something.

quickstarts-js/Get_Started.js
andycandy: Yes, we'll add "md" and "js" as classnames of output cells to classify them consistently.

quickstarts-js/Get_Started.js
andycandy: Added

quickstarts-js/Get_Started.js
andycandy: I have switched all proxy links to normal ones.

quickstarts-js/Get_Started.js
andycandy: Ok, I'll try to come up with something.

quickstarts-js/Get_Started.js
andycandy: Changed the example to pirate one.

quickstarts-js/Get_Started.js
```diff
@@ -1,158 +1,1098 @@
...
+// [CODE ENDS]
+
+/* Markdown (render)
+## Send text prompts
+
+Use the `generateContent` method to generate responses to your prompts. You can pass text directly to `generateContent` and use the `.text` property to get the text content of the response. Note that the `.text` field will work when there's only one part in the output.
+*/
+
+// [CODE STARTS]
+resp = await ai.models.generateContent({
```
Giom-V, line 70: You forgot to update that response

quickstarts-js/Get_Started.js
```diff
@@ -1,158 +1,1098 @@
...
+/* Markdown (render)
+### Upload text file
+
+Let's start by uploading a text file. In this case, you'll use a 400 page transcript from [Apollo 11](https://www.nasa.gov/history/alsj/a11/a11trans.html).
+*/
+
+// [CODE STARTS]
+TEXT_URL = "https://storage.googleapis.com/generativeai-downloads/data/a11.txt"
+
+resp = await fetch(TEXT_URL);
```
Giom-V, line 779: This one as well should be "response"

quickstarts-js/Get_Started.js
```diff
@@ -1,158 +1,1098 @@
...
+## Save and resume a chat
+
+In the JS SDK, chat history is represented as a plain array of messages, making it easy to serialize and resume sessions.
+
+#### 1. Save the chat history
+*/
+
+// [CODE STARTS]
+chatHistory = chat.getHistory()
+// [CODE ENDS]
```
Giom-V, line 322: No output here?

Giom-V: @andycandy to be fixed in a fast follow

quickstarts-js/Get_Started.js
Giom-V: @andycandy to be fixed in a fast follow

quickstarts-js/Get_Started.js
Giom-V: @andycandy to be fixed in a fast follow

quickstarts-js/Get_Started.js
andycandy: Yes, there is no output in this one. Do you want me to log the content of chatHistory?

PR #854, opened 2025-07-14
Added "JSON_mode" notebook for javascript

andycandy:


quickstarts-js/JSON_mode.js
```diff
@@ -0,0 +1,194 @@
...
+# Gemini API: JSON Mode Quickstart
+
+
+The Gemini API can be used to generate a JSON output if you set the schema that you would like to use.
+
+Two methods are available. You can either set the desired output in the prompt or supply a schema to the model separately.
+
+
+## Setup
+### Install SDK and set-up the client
```
Giom-V: Two level 3 titles without anything in between

quickstarts-js/JSON_mode.js
```diff
@@ -0,0 +1,194 @@
...
+// [CODE STARTS]
+module = await import("https://esm.sh/@google/genai@1.4.0");
+GoogleGenAI = module.GoogleGenAI;
+ai = new GoogleGenAI({ apiKey: process.env.GEMINI_API_KEY });
+
+MODEL_ID = "gemini-2.5-flash" // ["gemini-2.5-flash-lite-preview-06-17", "gemini-2.5-flash", "gemini-2.5-pro"]
+// [CODE ENDS]
+
+/* Markdown (render)
+## Set your constrained output in the prompt
```
Giom-V: I think you should first explain what Json mode is.

quickstarts-js/JSON_mode.js
```diff
@@ -0,0 +1,194 @@
...
+prompt = `
+  List a few popular cookie recipes using this JSON schema:
+
+  Recipe = {'recipe_name': str}
+  Return: list[Recipe]
+`
+// [CODE ENDS]
+
+/* Markdown (render)
+Now select the model you want to use in this guide, either by selecting one in the list or writing it down. Keep in mind that some models, like the 2.5 ones are thinking models and thus take slightly more time to respond (cf. [thinking notebook](./Get_started_thinking.ipynb) for more details and in particular learn how to switch the thiking off).
```
Giom-V: Wrong link (point to the python cookbook for now as it will be easier to find them all and fix them)

quickstarts-js/JSON_mode.js
```diff
@@ -0,0 +1,194 @@
...
+
+For this first example just describe the schema you want back in the prompt:
+
+*/
+
+// [CODE STARTS]
+prompt = `
+  List a few popular cookie recipes using this JSON schema:
+
+  Recipe = {'recipe_name': str}
```
Giom-V: I think you should also explain where the format comes from since it's a python thing if I'm not mistaken.

quickstarts-js/JSON_mode.js
```diff
@@ -0,0 +1,194 @@
...
+  }
+] 
+```
+
+*/
+
+/* Markdown (render)
+## Supply the schema to the model directly
+
+The newest models (1.5 and beyond) allow you to pass a schema object (or a python type equivalent) directly and the output will strictly follow that schema.
```
Giom-V: I think we can drop the "newest models" as the older ones are not available anymore.

quickstarts-js/JSON_mode.js
```diff
@@ -0,0 +1,194 @@
...
+
+console.log("```\n",rawResponse.text,"\n```")
+// [CODE ENDS]
+
+/* Output Sample
+
+```
+ [
+  {
+    &quot;recipe_name&quot;: &quot;Chocolate Chip Cookies&quot;
```
Giom-V: I think there's an encoding error here.

quickstarts-js/JSON_mode.js
```diff
@@ -0,0 +1,194 @@
...
+
+### Related examples
+
+* The constrained output is used in the [Text summarization](https://github.com/google-gemini/cookbook/blob/main/examples/json_capabilities/Text_Summarization.ipynb) example to provide the model a format to summarize a story (genre, characters, etc...)
+* The [Object detection](https://github.com/google-gemini/cookbook/blob/main/examples/Object_detection.ipynb) examples are using the JSON constrained output to uniiformize the output of the detection.
+
+### Continue your discovery of the Gemini API
+
+JSON is not the only way to constrain the output of the model, you can also use an [Enum](https://github.com/google-gemini/cookbook/blob/main/quickstarts/Enum.ipynb). [Function calling](https://github.com/google-gemini/cookbook/blob/main/quickstarts/Function_calling.ipynb) and [Code execution](https://github.com/google-gemini/cookbook/blob/main/quickstarts/Code_Execution.ipynb) are other ways to enhance your model by either using your own functions or by letting the model write and run them.
```
Giom-V: Maybe add a link to the quickstart as that's the only one we have at the moment.

quickstarts-js/JSON_mode.js
Giom-V: Check how we do it in the documentation, but µI think we should explain why you would need to use json mode (like an example), and then how it works.

Giom-V (changes requested):

andycandy:
@Giom-V Resolved all issues. PTAL

PR #866, opened 2025-07-21
Return Markdown in the web browser example

markmcd:
Currently the visual tool returns an image but if we return markdown too, it can also browse links on the page, and grab any text that is off-screen.


markmcd:
lint failure is due to "native", which is part of the model name. False positive.

PR #868, opened 2025-07-22
fix flash-lite model name

nikitamaia:
update flash lite model string in get started notebook


lucianommartins (approved):
LGTM, approved.

PR #892, opened 2025-08-07
Updated Get_started_LiveAPI.ipynb model to `gemini-live-2.5-flash-preview`

ROHANDWIVEDI2005:
- replaced all instances of gemini 2.0 to 2.5 
- rerun the notebook
- nbfmt


Giom-V:
@ROHANDWIVEDI2005 Thanks for the submission but I just saw that you left you API key in clear in the submission, which is why I'm closing the PR right away. I think the harm has already been done, and you need to deactivate it and create a new one ASAP as it will still be in Github's history.

As for the submission in itself. I'm not a fan of using preview models by default in our samples and they could change or be deprecated. The best would be, as Gemini was also recommending, to use a model selector so you can select either the official model or one of the preview ones:
`MODEL_ID = "gemini-live-2.5-flash-preview" # @param ["gemini-2.0-flash-live-001", "gemini-live-2.5-flash-preview","gemini-2.5-flash-preview-native-audio-dialog"] {"allow-input":true, isTemplate: true}`

Can you do that?

ROHANDWIVEDI2005:
> @ROHANDWIVEDI2005 Thanks for the submission but I just saw that you left you API key in clear in the submission, which is why I'm closing the PR right away. I think the harm has already been done, and you need to deactivate it and create a new one ASAP as it will still be in Github's history.
> 
> As for the submission in itself. I'm not a fan of using preview models by default in our samples and they could change or be deprecated. The best would be, as Gemini was also recommending, to use a model selector so you can select either the official model or one of the preview ones: `MODEL_ID = "gemini-live-2.5-flash-preview" # @param ["gemini-2.0-flash-live-001", "gemini-live-2.5-flash-preview","gemini-2.5-flash-preview-native-audio-dialog"] {"allow-input":true, isTemplate: true}`
> 
> Can you do that?

@Giom-V  Thanks for the save this is strange I did remove the API key but somehow it still showed in the pr . I guess I should have checked after the pr was made I guess I will use a .env file for this in the future

PR #903, opened 2025-08-18
Adding to prompting examples

wd021:
Adding some helpful prompt resources. [God Tier Prompts](https://www.godtierprompts.com) is a community leaderboard users can use to discover and find the best prompts.


Giom-V:
@wd021 Thank you for you PR, but I think it's too early to mention this godtierprompts website as at the moment it has vey few content for Gemini users.

Issue #911, opened 2025-08-25
AI-powered GitHub Action


ROHANDWIVEDI2005:
### Description of the feature request:

This is a proposal to add a new cookbook example: an AI-powered GitHub Action that converts plain text into well-structured Markdown documentation.

### What problem are you trying to solve with this feature?

Many developers need to create clear documentation (like `README.md` files) but find manual formatting tedious. This tool automates the process by leveraging the Gemini API to intelligently structure the text, saving time and ensuring consistency.

The workflow is triggered by a push to a source text file (e.g., `source.txt`). It then:
1. Reads the content of the text file.
2. Calls the Gemini API to generate a Markdown version of the text.
3. Automatically commits the new Markdown content to a destination file (e.g., `README.md`).

This makes the documentation process a simple "write and push" action.

The complete and working code for this example is available in my public repository:

[**https://github.com/ROHANDWIVEDI2005/Convert_to_md**]

<img width="928" height="896" alt="Image" src="https://github.com/user-attachments/assets/df89dd04-0458-4921-a95a-425d8f16bac7" />

ROHANDWIVEDI2005:
#915

Giom-V:
Thanks for the PR @ROHANDWIVEDI2005, but I have to say I'm not sure of the value added for the cookbook. If it's just a text to markdown converter, maybe just a notebook example would be enough, no?

ROHANDWIVEDI2005:
Thanks for your feedback @Giom-V  but the aim for this example was to showcase how the gemini-api capabilities can be used in a real world CI/CD environment  **A github action which can automate a task**  ,but i guess I could  add a companion notebook showcasing a more detailed step-by-step walk-through of the example.

ROHANDWIVEDI2005:
@Giom-V  If an end-to-end working example is not suitable for the cookbook would it be better to create a notebook about the same  markdown-to-text converter but  implementing it with the  `PyGithub`  library instead of the GitHub-actions approach It will be similar to #887 and still retaining the purpose of task automation.

ROHANDWIVEDI2005:
@markmcd is this something that would be usefull to the gemini-cli repo

PR #915, opened 2025-08-26
added example of github actions

ROHANDWIVEDI2005:
- Added an example of a txt to markdown converter github action
- Demonstrates the use of gemini api in github-action work flows.


ROHANDWIVEDI2005:
Resolves issue #911

PR #928, opened 2025-09-04
feat: Add test coverage for file-api example

Mwessc:
Refactored the `quickstarts/file-api/sample.py` script into functions to make it testable.

Added a new test file `quickstarts/file-api/test_sample.py` with unit tests for the refactored functions.

The new tests use `unittest.mock` to mock the Google API client and file operations, allowing the tests to run without any external dependencies.

The tests cover the following functionality:
- File uploading
- File retrieval
- Content generation with a file
- File deletion

This change significantly improves the test coverage of the `quickstarts/file-api` example.


Giom-V:
@Mwessc, Thanks for the submission but I don't think we want to add test coverage to the cookbook. The cookbook is meant to showcase easy to understand code snippets, so adding tests on top of them would only make them more complex for new users.

PR #943, opened 2025-09-09
update batch api cookbook

patrickloeber:


quickstarts/Batch_mode.ipynb
```diff
@@ -54,7 +54,10 @@
         "3.  Create a batch job using inline requests (convenient for smaller jobs).\n",
         "4.  Monitor the status of your job.\n",
         "5.  Retrieve and parse the results for both job types.\n",
-        "6.  Manage your jobs (list and cancel)."
+        "6.  Manage your jobs (list and cancel).\n",
+        "7.  Use batch embeddings\n",
+        "8.  Multimodal input\n",
+        "9.  Image generation"
```
markmcd: Nit: The old ones are all sentences, but the new ones are fragments. How about something like this?

```suggestion
        "7.  Use batch embeddings.\n",
        "8.  Generate with multimodal inputs.\n",
        "9.  Generate images."
```

quickstarts/Batch_mode.ipynb
```diff
@@ -777,6 +780,167 @@
...
+            ]
+          },
+          "execution_count": 10,
+          "metadata": {},
+          "output_type": "execute_result"
+        }
+      ],
+      "source": [
+        "uploaded_batch_requests = client.files.upload(\n",
+        "    file=\"embeddings_batch_requests.jsonl\", config={\"mime_type\": \"jsonl\"}\n",
```
markmcd: Do we need the mime type? The other examples don't use it.

(if we do, we should use the proto-style syntax - `types.UploadFileConfig(mime_type=...)`, see below)

quickstarts/Batch_mode.ipynb
```diff
@@ -777,6 +780,167 @@
...
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {
+        "id": "c4970750f9e9"
+      },
+      "outputs": [],
+      "source": [
+        "batch_job = client.batches.create_embeddings(\n",
+        "    model=\"gemini-embedding-001\",\n",
+        "    src={'file_name': uploaded_batch_requests.name}\n",
```
markmcd:
```suggestion
        "    src=types.EmbeddingsBatchJobSource(file_name=uploaded_batch_requests.name)\n",
```

tbh I don't mind using dicts but Ali was asking to keep samples aligned on this proto-style syntax

quickstarts/Batch_mode.ipynb
```diff
@@ -777,6 +780,167 @@
...
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {
+        "id": "165cf44fa5ce"
+      },
+      "outputs": [],
+      "source": [
+        "while True:\n",
+        "    batch_job = client.batches.get(name=batch_job.name)\n",
+        "    if batch_job.state.name in ('JOB_STATE_SUCCEEDED', 'JOB_STATE_FAILED', 'JOB_STATE_CANCELLED'):\n",
```
markmcd, line 947: Optional, but are there any other states that might exist? Would this be better testing like so?

```py
if batch_job.state.name != 'JOB_STATE_PENDING':
```

quickstarts/Batch_mode.ipynb
```diff
@@ -120,7 +123,7 @@
       "source": [
         "### Initialize SDK client\n",
         "\n",
-        "With the new SDK you now only need to initialize a client with you API key (or OAuth if using [Vertex AI](https://cloud.google.com/vertex-ai)). The model is now set in each call."
+        "Initialize a client with you API key"
```
markmcd:
```suggestion
        "Initialize a client with your API key"
```

markmcd (approved):
Left a few nits. I can make these changes directly if you prefer?

quickstarts/Batch_mode.ipynb
patrickloeber: yes, needed for jsonl, and good practice to also use it for json

PR #990, opened 2025-10-11
Migrating OAuth notebook

Aarchi-07:
- Updated SDK imports
- lint checked
- nbfmt

### DIFF
--- a/quickstarts/Authentication_with_OAuth.ipynb
+++ b/quickstarts/Authentication_with_OAuth.ipynb
@@ -177,8 +177,7 @@
       "cell_type": "code",
       "execution_count": null,
       "metadata": {
-        "id": "uRg4GMDQLPKl",
-        "outputId": "b0883075-5217-40e6-b444-27e67f0d4d45"
+        "id": "uRg4GMDQLPKl"
       },
       "outputs": [
         {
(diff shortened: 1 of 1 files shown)


Aarchi-07:
I was unable to migrate the "Authenticate with the Python SDK by service account key" section and would appreciate any guidance.

quickstarts/Authentication_with_OAuth.ipynb
```diff
@@ -284,15 +285,27 @@
...
+        }
+      ],
       "source": [
-        "import google.generativeai as genai\n",
+        "from google import genai\n",
+        "from google.colab import userdata\n",
         "\n",
-        "print('Available base models:', [m.name for m in genai.list_models()])"
+        "GOOGLE_API_KEY=userdata.get('GOOGLE_API_KEY')\n",
+        "client = genai.Client(api_key=GOOGLE_API_KEY)\n",
```
kkorpal, line 306: Just a quick fix needed here. Since this is the OAuth Authentication tutorial, passing api_key=GOOGLE_API_KEY actually bypasses the OAuth setup you just taught the user and uses the standard API key instead.
To properly test that the OAuth credentials are working, we need to initialize an empty client so the SDK picks them up automatically in the background. Could you update?

kkorpal (changes requested):

PR #998, opened 2025-10-16
Implement JS version of shape sequence prediction

Aarchi-07:
- Converted Python “guess the shape” logic to JavaScript.
- Added functions to fetch and display images in Base64.
- Integrated prompt with AI model to predict the next shape.


kkorpal:
Hi @Aarchi-07, thanks for porting another great example to JS!
Just like your other PR, the automated header-check is failing here. Could you please add the standard Apache 2.0 License header block to the very top of your new file?
let us know once the license check is green!

Aarchi-07:
@kkorpal Updated, PTAL!

Aarchi-07:
@kkorpal , please review again.

kkorpal:
Hi @Aarchi-07, thanks for porting this! To make it run on our interactive player, please:
1. Add `const` to all variable declarations to avoid strict-mode ReferenceErrors.
2. Attach shared variables (like `triangleBase64` and `prompt`) to `globalThis` so separate blocks can access them.
3. Remove the unawaited `(async () => { ... })()` wrapper around the image fetch to prevent a race condition before the model runs.

Thanks.!

PR #1002, opened 2025-10-20
Colab update for robotics

PaulTR:


Giom-V:
Considering the length of the notebook, we should add a table of content and `<a name="foo" />` hidden links. Check how Mark and I did in on his new datasets notebook.
Also you'll need to find a way to keep it under 19M...

PaulTR:
So we ended up with a robotics-samples repo. I think I'm going to break this one into multiple smaller colabs there, then use this as a table of contents.

PR #1020, opened 2025-10-31
`GOOGLE_API_KEY` -> `GEMINI_API_KEY`

markmcd:
Colab now uses `GEMINI_API_KEY` for keys imported from AI Studio, along with the default across the docs on ai.google.dev.

This brings the cookbook in line with the same defaults so users don't need to figure out they have to rename the key.


Giom-V:
I'm all for it, but what about those who manually copied their API key under the name "GOOGLE_API_KEY"?

Giom-V:
I think we should add a "breaking change" on the main readme (and maybe remove the migration guide now that it's been nearly a year). Also I think CONTRIBUTING.md and .gemini/styleguide.md should also be updated accordingly. And maybe the authentication notebook should also be updated with explanations about the auto-import feature.

Giom-V:
I'm going to work on a new PR that's going to be easier

PR #1075, opened 2025-12-19
Fix #906: Add websocket tool response examples and documentation

### LINKED ISSUE #906: Gemini Live API - Websocket Tool Call Response - BidiGenerateContentToolResponse
### Description of the bug:

Gemini Live API, Websockets function calling

The docs are unclear about what BidiGenerateContentToolResponse is like. it says that the field is functionResponses[], but when I send that it does not work, there is no documentaiton on what it should look like

After a tool call, the docs say I think I need to send a BidiGenerateContentToolResponse to the weboscket. I do not know the format, and the docs are unclear. Any help would be appreciated, thanks!

The cookbook here used this, but its not for websockets:
function_response = types.FunctionResponse(
id=fc.id,
name=fc.name,
response={“result”: “ok”},
)
https://colab.research.google.com/github/google-gemini/cookbook/blob/main/quickstarts/Get_started_LiveAPI_tools.ipynb#scrollTo=EmTKF_DtrY4U

Im trying different variations of this:
tool_response_payload = {
“functionResponse”: {
“functionCallId”: tool_call_id,
“functionName”: tool_name,
}
}

await ws.send(json.dumps(tool_response_payload))

https://ai.google.dev/api/live#FunctionResponse

_No response_

ayushdebnath012:
Hey! This fixes issue #906.

I ran into the same problem many others reported: the docs mention `BidiGenerateContentToolResponse`, but they don’t clearly show how to format a tool response when using WebSockets. I tried several approaches, but the WebSocket kept closing unexpectedly.

WHAT I ADDED

Two files to make WebSocket function calling actually usable end-to-end:

1) Websocket_Tool_Call_Response_Example.py  
A complete, runnable example. You just need to add your API key.
It shows:
- How to set up a WebSocket connection with tools enabled
- How to detect and handle a function call
- Two different (working) ways to send the tool response back
- Example tools (weather lookup, turning lights on/off)

2) WEBSOCKET_TOOL_RESPONSE_FIX.md  
A practical guide covering:
- What works and what doesn’t
- Common mistakes that cause WebSocket disconnects
- Subtleties missing from the current docs

THE ACTUAL SOLUTION

There are two formats that work for sending tool responses over WebSockets.

OPTION 1 (used in the cookbook examples)

    response_msg = {
        "tool_response": {
            "function_responses": [{
                "id": function_call_id,
                "name": function_name,
                "response": {"result": your_data}
            }]
        }
    }

OPTION 2 (based on @nmfisher’s workaround in the issue)

    response_msg = {
        "clientContent": {
            "turns": [{
                "role": "user",
                "parts": [{
                    "functionResponse": {
                        "name": function_name,
                        "response": {"output": your_data}
                    }
                }]
            }],
            "turnComplete": True
        }
    }

Both formats work reliably. I’ve been using Option 1 since it matches the cookbook examples.

WHAT DOES NOT WORK

The following consistently caused failures or WebSocket closures:
- Trying to use BidiGenerateContentToolResponse directly
- Forgetting the function call ID
- Not matching the function name exactly
- Sending a Python dict instead of a JSON-encoded string

TESTED SCENARIOS

Tested with:
- Single function calls
- Multiple function calls in a single turn
- Follow-up model responses after a function call

### DIFF
--- a/examples/Websocket_Tool_Call_Response_Example.py
+++ b/examples/Websocket_Tool_Call_Response_Example.py
@@ -1,0 +1,351 @@
+# -*- coding: utf-8 -*-
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Example: Handling Tool Call Responses with Gemini Live API Websockets
+
+This example demonstrates the CORRECT way to send function/tool call responses
+back to the Gemini Live API using websockets.
+
+## The Issue (GitHub #906)
+The documentation mentions BidiGenerateContentToolResponse, but the actual
+format that works is different. This example shows the working solution.
+
+## Setup
+
+Install dependencies:
+```
+pip install websockets
+```
+
+Set your API key:
+```
+export GOOGLE_API_KEY="your-api-key"
+```
+
+## Run
+```
+python Websocket_Tool_Call_Response_Example.py
+```
+"""
+
+import asyncio
+import json
+import os
+from websockets.asyncio.client import connect
+
+# Configuration
+GOOGLE_API_KEY = os.environ.get("GOOGLE_API_KEY")
+if not GOOGLE_API_KEY:
+    raise ValueError("GOOGLE_API_KEY environment variable must be set")
+
+host = "generativelanguage.googleapis.com"
+model = "gemini-2.0-flash-live-001"
+uri = f"wss://{host}/ws/google.ai.generativelanguage.v1beta.GenerativeService.BidiGenerateContent?key={GOOGLE_API_KEY}"
+
+
+# Example tool functions
+def get_weather(location: str) -> dict:
+    """Mock function to get weather information."""
+    return {
+        "location": location,
+        "temperature": "72°F",
+        "condition": "Sunny",
+        "humidity": "45%"
+    }
+
+
+def turn_on_lights() -> dict:
+    """Mock function to turn on lights."""
+    return {"status": "success", "message": "Lights turned on"}
+
+
+def turn_off_lights() -> dict:
+    """Mock function to turn off lights."""
+    return {"status": "success", "message": "Lights turned off"}
+
+
+# Tool registry mapping function names to implementations
+TOOL_REGISTRY = {
+    "get_weather": get_weather,
+    "turn_on_lights": turn_on_lights,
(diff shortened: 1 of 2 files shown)


ayushdebnath012:
@bebechien can we merge this pr plz

examples/Websocket_Tool_Call_Response_Example.py
```diff
@@ -0,0 +1,363 @@
...
+
+# Tool registry mapping function names to implementations
+TOOL_REGISTRY = {
+    "get_weather": get_weather,
+    "turn_on_lights": turn_on_lights,
+    "turn_off_lights": turn_off_lights,
+}
+
+
+async def setup_session(ws, model="gemini-2.0-flash-live-001"):
```
kkorpal, line 88: Please updated to a currently supported model

examples/Websocket_Tool_Call_Response_Example.py
```diff
@@ -0,0 +1,363 @@
...
+    """Initialize the websocket session with model configuration and tools."""
+    setup_msg = {
+        "setup": {
+            "model": f"models/{model}",
+            "generation_config": {
+                "response_modalities": ["TEXT"]
```
kkorpal, line 94: native audio models don't support TEXT modality. This causes an immediate ConnectionClosedError on setup. It should be ["AUDIO"], or the example should use a model that supports TEXT.

examples/Websocket_Tool_Call_Response_Example.py
```diff
@@ -0,0 +1,363 @@
...
+            ]
+        }
+    }
+    
+    print("📤 Sending setup message...")
+    await ws.send(json.dumps(setup_msg))
+    
+    # Wait for setup confirmation
+    raw_response = await ws.recv(decode=False)
+    setup_response = json.loads(raw_response.decode("ascii"))
```
kkorpal, line 132: This crashes with UnicodeDecodeError when responses contain non-ASCII characters (e.g., ° in 72°F). Should be "utf-8"

examples/Websocket_Tool_Call_Response_Example.py
```diff
@@ -0,0 +1,363 @@
...
+async def main(model="gemini-2.0-flash-live-001"):
+    """Main function demonstrating tool call handling."""
+    print("=" * 70)
+    print("Gemini Live API - Websocket Tool Call Response Example")
+    print(f"Using model: {model}")
+    print("=" * 70)
+    
+    async with connect(
+        uri, 
+        additional_headers={"Content-Type": "application/json"}
```
kkorpal, line 314: The default is 10 seconds, which can be too short on slow networks. \> Use async with connect(uri, additional_headers={"Content-Type": "application/json"}, open_timeout=30) as ws: to avoid timeout errors.

kkorpal (changes requested):

kkorpal:
Hi @ayush31010, Thanks for your contribution! I will bring this to the engineering team for review. In the meantime, I've added some inline comments; please take a look at those and ensure the code and documentation  follow our [Cookbook Style Guide](https://github.com/google-gemini/cookbook/blob/main/CONTRIBUTING.md#style-guides), which aligns with the [Google Developer Style Guide](https://developers.google.com/style). 
 
Thanks..!

PR #1189, opened 2026-04-04
fix api key inconsistency and cleanup in live api quickstart

### LINKED ISSUE #948: Improve Gemini Live API Cookbook
### Description of the bug:

Recommendations below are relevant to this Colab notebook : https://github.com/google-gemini/cookbook/blob/main/quickstarts/Get_started_LiveAPI.ipynb

1. fix the Colab GOOGLE_API_KEY

```python
from google.colab import userdata
import os

os.environ['GOOGLE_API_KEY'] = userdata.get('GOOGLE_API_KEY')
```

to

```python
from google.colab import userdata
import os

GOOGLE_API_KEY = userdata.get('GOOGLE_API_KEY')
```

2. The colab sample does not support `gemini-2.5-flash-preview-native-audio-dialog` .

when selecting 
```python
MODEL = "gemini-2.5-flash-preview-native-audio-dialog"
```

then the `Text to Text` sample run into an error

3. the section `Working with resumable sessions` does not reuse the `client` and `model` defined at the beginning of the colab.

1. actual behavior : the code below fails to execute. expected behavior : either use the GOOGLE_API_KEY var or remove the argument and use OS env.

```python
client = genai.Client(api_key=GOOGLE_API_KEY)
```

2.  actual behavior : the code samples with TEXT modalities fail to execute when `gemini-2.5-flash-preview-native-audio-dialog` is selected. expected behavior : instruct the user that TEXT modalities is only supported by non-native audio model or update the native audio API to support TEXT modality.

```python
config={
    "response_modalities": ["TEXT"]
}
```

3.  actual behavior : model & client variables are redefined. expected behavior :  remove or comments the following lines from the section `Working with resumable sessions > Helper functions`: 

```python
MODEL =  "gemini-live-2.5-flash-preview"

client = genai.Client(api_key=GOOGLE_API_KEY)
```

_No response_

knQzx:
## Summary

- fixed undefined GOOGLE_API_KEY variable — now reads from os.environ like the rest of the notebook
- added a warning when native-audio model is selected since it doesn't support TEXT response modality
- removed redundant client/model redefinition in the resumable sessions section

## Test plan

- [x] verified notebook cells reference the correct api key variable
- [x] verified model guard prints warning for native-audio models
- [x] verified resumable sessions section works without redundant client init

fixes #948


knQzx:
I have signed the CLA

Venkaiahbabuneelam:
Hi @knQzx, Thanks for reaching out us!

Regarding you PR few checks were failing, Could you please have a look once.

Thanks

knQzx:
pushed nbfmt formatting only - content is identical just whitespace so notebook format should pass

notebook lint will still fail on google::inclusive_language - it flags the word native in the model id gemini-2.5-flash-native-audio-preview-09-2025 which is a product name and can't be renamed

same thing happens on main for this file - lint just doesn't surface there because it only runs on notebooks touched by a pr - 5 other notebooks in main also reference native-audio (Browser_as_a_tool, LiveAPI_plotting_and_mapping, Get_started_LiveAPI_tools, Models, Search_Grounding)

the native-audio guard is literally the fix for #948 so removing it to please the linter doesn't make sense

on your side you could add --exclude_lint=google::inclusive_language to the workflow or whitelist product ids in the wordlist

also the run is stuck in action_required - could you approve it

quickstarts/Get_started_LiveAPI.ipynb
```diff
@@ -160,16 +160,12 @@
...
         "id": "HghvVpbU0Uap"
       },
       "outputs": [],
-      "source": [
-        "from google import genai\n",
-        "from google.genai import types\n",
-        "client = genai.Client(api_key=GOOGLE_API_KEY)"
-      ]
+      "source": "from google import genai\nfrom google.genai import types\nclient = genai.Client(api_key=os.environ.get('GOOGLE_API_KEY'))"
     },
```
kkorpal, line 170: When I run this cell in Colab, it throws a NameError: name 'GOOGLE_API_KEY' is not defined. We need to fetch the key from the environment.

kkorpal (changes requested):

thorwebdev:
@knQzx this is great, thanks so much. Could you also update the `send_client_content` to use `await session.send_realtime_input(text="Hello, how are you?")` instead and use `gemini-3.1-flash-live-preview` as the default? https://ai.google.dev/gemini-api/docs/live-api/capabilities#send-text

knQzx:
@thorwebdev done in 161c951

- replaced send_client_content with send_realtime_input(text=...) in all 4 call sites (Text to Text, Simple text to audio, AudioLoop.send, resumable sessions send)
- default MODEL bumped to gemini-3.1-flash-live-preview, kept the other ids in the dropdown so users can still pick them

native-audio guard left in place - now only fires if someone manually picks the native-audio model from the dropdown

thorwebdev:
Hey @knQzx I'm sorry, I should have mentioned that `gemini-3.1-flash-live-preview` is also a native audio model and also does not support `"response_modalities": ["TEXT"],` only `AUDIO`, so this would need a bit of a larger refactor. 

Would you be up for changing every cell to AUDIO by any chance and removing the initial Text to Text cell, seeing that the only model that used to support this has since been deprecated? https://ai.google.dev/gemini-api/docs/deprecations#live-api-models

Sorry about that!

thorwebdev:
@knQzx I've made the update here now: https://github.com/google-gemini/cookbook/pull/1222

I really appreciate you getting started on these changes and contributing your time!

knQzx:
> @knQzx I've made the update here now: #1222
> 
> I really appreciate you getting started on these changes and contributing your time!

thanks @thorwebdev, all good. closin this in favor of #1222 which has the broader cleanup anyway. happy to help on the next one!!

PR #1211, opened 2026-05-01
quickstarts/Embeddings.ipynb: fix variable name typo and truncate vector print outputs

ShriPunta:
Two readability fixes in `Embeddings.ipynb` :

1. Renamed `GEMINI_EMBEDDIG_1_MODEL_ID` → `GEMINI_EMBEDDING_1_MODEL_ID` (missing 'N') in 3 places.
2. Truncated full 3072-element embedding vector prints in 4 cells to a 50-character preview with `... TRIMMED]`, consistent with the existing style used elsewhere in the notebook. Also cleared the stored cell outputs for those cells.

While viewing the file on GitHub the printed embedding array output blows up the file size.


kkorpal:
Hi @ShriPunta, thank you for fixing that typo and truncating the massive embedding outputs! It definitely makes the file much easier to read on GitHub.
It looks like the Notebooks / Notebook lint check is failing. This is likely because you cleared the stored cell outputs. Our repository requires notebooks to be saved with their execution outputs visible so readers can see the expected results without having to run the code themselves.
Could you please run the notebook from top to bottom so your new truncated prints are generated, save it, and push the updated file? Let us know once the checks turn green so we can move a head!

PR #1236, opened 2026-05-22
Fix scale= command in Lyria realtime quickstart

aryanmotgi:
## Bug

In `quickstarts/Get_started_LyriaRealTime.py`, the `scale=` command parser compares the raw `prompt_str` (which still contains the `scale=` prefix, e.g. `scale=A_FLAT_MAJOR_F_MINOR`) against `types.Scale` enum names (e.g. `A_FLAT_MAJOR_F_MINOR`). The equality check never succeeds and the command always falls through to `Error: Matching enum not found.`, making the `scale=` command effectively non-functional.

## Fix

Strip the `scale=` prefix (and whitespace) before comparing, mirroring how the sibling `bpm=` (L121) and `top_k=` (L148) handlers already work.

## Repro

```
> scale=A_FLAT_MAJOR_F_MINOR
Error: Matching enum not found.
```

After the fix:
```
> scale=A_FLAT_MAJOR_F_MINOR
Setting scale to A_FLAT_MAJOR_F_MINOR, which requires resetting context.
```

### DIFF
--- a/quickstarts/Get_started_LyriaRealTime.py
+++ b/quickstarts/Get_started_LyriaRealTime.py
@@ -130,9 +130,10 @@ async def main():
                     del config.scale
                     print(f"Setting Scale to AUTO, which requires resetting context.")
                   else:
+                    scale_name = prompt_str.removeprefix('scale=').strip()
                     found_scale_enum_member = None
                     for scale_member in types.Scale: # types.Scale is an enum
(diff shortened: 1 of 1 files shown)


kkorpal:
Hi @aryanmotgi, Thank you for catching and fixing this bug in the Lyria quickstart! We will review and let you know. Thanks.

Giom-V (approved):

PR #1258, opened 2026-06-09
Add service account auth section with IAM roles for Gemini, File, and…

### LINKED ISSUE #1254: Authentication guide needs to be updated with API key and Service Account combo
### Description of the bug:

Currently obtaining and API key is a problem, as its not documented what permissions you need to setup for service account key file, which is now mandatory. 

Improve the Authentication guide, and explain what IAM permissions/roles need to be added to the binded service account key entry in order to use Gemini and File Search API and other connected services, which need a lot more now than just grab an api key. 

_No response_

kkorpal:
… Vertex AI APIs , Fix #1254


quickstarts/Authentication.ipynb
```diff
@@ -335,6 +335,164 @@
         "## Store your key in an environment variable"
       ]
     },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "28f8532108d0"
+      },
+      "source": [
+        "## Authenticate with a service account\n",
```
Giom-V: Just add links about Vertex, no details as we need to focus on API keys

Giom-V (approved):

PR #1301, opened 2026-07-28
Fix malformed JSON in examples/gemini_google_adk_model_guardrails.ipynb

geyee:
The notebook `examples/gemini_google_adk_model_guardrails.ipynb` fails to parse as JSON and cannot be opened in Google Colab. One cell source array string element is missing its trailing comma (at char 20561, line 559). This PR adds the single missing comma so the file is valid JSON again, with no other content changes (+1 byte total). After the fix the notebook parses cleanly with json.load and opens in Colab.


kkorpal:
Hi @geyee ! The Google CLA check is failing because it's missing a signed Contributor License Agreement.

PR #1302, opened 2026-07-28
Remove unsupported presence_penalty/frequency_penalty from example

### LINKED ISSUE #1290: presence_penalty/frequency_penalty example uses 0.0 values that mask a 400 INVALID_ARGUMENT on every tested model
### Description of the bug:

In `quickstarts/Get_started_Generate_Content.ipynb`, the "Configure model parameters" section demonstrates:

```python
config=types.GenerateContentConfig(
    ...
    presence_penalty=0.0,
    frequency_penalty=0.0,
)
```

This executes successfully only because `0.0` is a no-op value that isn't validated server-side. Setting either parameter to any non-zero value (e.g. `0.2`) fails with:
```text
ClientError: 400 INVALID_ARGUMENT. {
  'error': {
    'code': 400,
    'message': 'Penalty is not enabled for this model',
    'status': 'INVALID_ARGUMENT'
  }
}
```
This is a known, cross-ecosystem issue, not specific to this notebook or the `google-genai` SDK:
- [Google AI Developer Forum: "Doom Loop - Frequency Penalty - 2.5 Flash"](https://discuss.ai.google.dev/t/doom-loop-frequency-penalty-2-5-flash/94446) — a Google team member confirms `frequencyPenalty` is only supported on `gemini-2.0-flash`, not on 2.5-series models.
- [vercel/ai#8943](https://github.com/vercel/ai/issues/8943) — same `Penalty is not enabled for models/gemini-2.5-*` error reproduced independently on `gemini-2.5-pro` and `gemini-2.5-flash`.
- [BerriAI/litellm#26108](https://github.com/BerriAI/litellm/issues/26108) — LiteLLM's own `get_supported_openai_params()` incorrectly lists `frequency_penalty` as supported for `gemini-2.5-flash`, and the API rejects it when actually used.

**Actual:** The notebook example runs without error because it uses `presence_penalty=0.0, frequency_penalty=0.0`, giving the false impression these parameters are functional. Using any non-zero value (the entire point of the parameter) fails with `400 INVALID_ARGUMENT` on every model I tested.

**Models tested (non-zero penalty values):**

| Model | Result |
|---|---|
| `gemini-3.5-flash` | ❌ `INVALID_ARGUMENT` — Penalty not enabled |
| `gemini-3.1-flash-lite` | ❌ `INVALID_ARGUMENT` — Penalty not enabled |
| `gemini-3.1-pro-preview` | ❌ `INVALID_ARGUMENT` — Penalty not enabled |
| `gemini-3-flash-preview` | ❌ `INVALID_ARGUMENT` — Penalty not enabled |
| `gemini-2.5-flash` | ❌ `INVALID_ARGUMENT` — Penalty not enabled for models/gemini-2.5-flash |

I wasn't able to complete testing on `gemini-2.5-pro`, `gemini-2.0-flash`, and `gemini-2.0-flash-lite` due to free-tier quota limits, but given the consistent pattern above, I suspect these are unsupported too.

**Expected:** Either the example should use values that actually demonstrate the parameter working (if supported on some model), or the parameters should be removed from the example if unsupported across the currently recommended models, so users aren't misled into thinking this is usable functionality.

Anyone copying this example and setting a real penalty value will hit a confusing, undocumented error. As the linked forum thread shows, this isn't just a notebook nitpick it's an actual blocking real production use cases (e.g. mitigating repetition/"doom loops").

Suggested fix (open to guidance on preferred approach):
- Remove `presence_penalty`/`frequency_penalty` from this example if they're not currently supported on any actively-recommended model, or
- If they are supported on some specific model/version not covered above, use a non-zero value in the example so it actually demonstrates working usage, with a note on which models support it.

Happy to submit a PR once there's agreement on which direction is preferred, per the CONTRIBUTING guidelines.

AryanSherigar:
## Description

Removes `presence_penalty` and `frequency_penalty` from the "Configure model parameters" example in `quickstarts/Get_started_Generate_Content.ipynb`.

These parameters were previously set to `0.0`, which silently masked the fact that they aren't supported on any currently recommended Gemini model. Setting either to a non-zero value (the entire point of the parameter) fails with:
```text
ClientError: 400 INVALID_ARGUMENT

{
  "error": {
    "code": 400,
    "message": "Penalty is not enabled for this model",
    "status": "INVALID_ARGUMENT"
  }
}
```
This is a known, cross-ecosystem issue and not specific to this notebook, as detailed in the linked issue.

## Related issue

Fixes #1290

## Changes

- Removed `presence_penalty=0.0` and `frequency_penalty=0.0` from the `GenerateContentConfig` example
- Re-ran the notebook cells to reflect the updated example and its output


kkorpal:
Thanks @AryanSherigar  for your contribution This is fixed in this [PR](https://github.com/google-gemini/cookbook/pull/1305). closing this PR.

PR #1323, opened 2026-08-14
fix: add disclaimer and generate_content fallback for video_metadata fps

kkorpal:
add disclaimer and generate_content fallback for video_metadata fps


Giom-V (changes requested):
Loooks good to me, but since you're updating this notebook, can you also change `nteraction.steps[-1].content[0].text` and 
```
text_output = next((o for o in interaction.steps if o.type == "text"), None)
if text_output:
    display(Markdown(text_output.text))
```
with `interaction.output_text`?

kkorpal:
Closing this PR as now interaction api is supporting FPS, its fixed in https://github.com/google-gemini/cookbook/pull/1335.

PR #1340, opened 2026-08-22
Migrate Role_prompting example to the Interactions API

ayushiyadav02:
Migrates `examples/prompting/Role_prompting.ipynb` from `client.models.generate_content()` to `client.interactions.create()`, per the Interactions API policy in CONTRIBUTING.md.

Changes:
- Both generation calls now use `client.interactions.create()` with `input=` and a top-level `system_instruction=`, replacing the `types.GenerateContentConfig` wrapper.
- Response access updated to `interaction.steps[-1].content[0].text`.
- Removed the now-unused `from google.genai import types` import.

No behavioral change — same model, prompts, and system instructions. Outputs re-run and included, matching the existing file.

Verified with nbfmt, nblint, and check_readme_links; all pass.

Happy to work through more of the examples/ notebooks still on `generate_content()` if that's useful — there are quite a few left.


ayushiyadav02:
Thanks for the review — two notes:

**Formatting / execution counts.** Re-ran `nbfmt`, which reports the notebook as already compliant (`Modified/Needs: 0, Errors: 0`). It did strip some leftover Colab `outputId` metadata, which I've pushed. `nbfmt` doesn't reset execution counts, and the version on `main` carries them as well, so I've left them as they are — happy to null them if you'd prefer.

**`MODEL_ID` line length.** That line is unchanged from `main` (byte-identical) and is the standard Colab form selector required by `gemini::model_selector`. `nblint` exempts form annotations from the line-length rule and doesn't flag it — the only warning it reports on this file is a pre-existing one in the tour-guide prompt. Shortening the selector would break it, so I've left it alone.

Let me know if you'd like anything adjusted.

kkorpal:
Hi @ayushiyadav02 , thanks for your contribution. It looks good to me, but since you're updating this notebook, just one quick tweak is needed: could you swap out `print(interaction.steps[-1].content[0].text)` for the cleaner `print(interaction.output_text)`?

kkorpal (approved):

PR #1376, opened 2026-09-13
Fix broken cells in the autonomous video notebook and lint for syntax errors

rishabhsai:
Cells 32, 33 and 34 of `examples/Autonomous_video_production_with_omni_and_agents.ipynb` do not compile: each assigns an embedded script with `generate_script_content = """#!/usr/bin/env python3`, and the script's own module docstring on the next line closes that literal, so `ast.parse` reports `invalid syntax` at line 2 of all three cells and none of `generate_script_content`, `review_script_content` or `extend_script_content` is ever bound. Cell 36, the managed agent launch, then fails with `NameError` on those three names plus `GEMINI_API_KEY`, which cell 10 puts in `os.environ` but never binds as a variable, so everything from "Create modular production scripts" to the end of the notebook is unreachable.

Switching the outer delimiter of the three assignments to `'''` fixes them without disturbing the embedded docstrings or the nested `f"""` prompt in `review_clip.py`, and cell 10 gains one line, `GEMINI_API_KEY = os.environ["GEMINI_API_KEY"]`, matching `quickstarts/Get_started_managed_agents.ipynb`.

To stop this recurring, `tools/nblint/rules/syntax.py` adds a `syntax::code_cell` rule that runs `ast.parse` over every code cell after neutralizing `%magic` and `!shell` lines and skipping `%%` cell magics, and reports the cell index and line of any `SyntaxError`; on the notebook as it stands on main it reports exactly cells 32, 33 and 34 at line 2, and across all 150 notebooks in the repo it reports nothing else.

Verified with `python3 tools/nblint_cli.py --repo=google-gemini/cookbook --branch=main` over every notebook (150 checked, 0 failed) plus `nbfmt_cli.py --test` and `check_readme_links.py` on the changed file; the three cells were also run in a plain Python process and reproduce their committed output exactly, so no output or metadata is touched and the notebook was not re-executed end to end, which would need an API key. The cells came in with #1369 through #1372 on 2026-09-09.


rishabhsai:
Looked at the three Gemini Code Assist suggestions. The import order and the `cell.get("source", [])` pattern match every existing rule module in tools/nblint/rules (structure.py, gemini.py, style.py), so I kept the new rule consistent with those. When `e.text` is empty the message already carries `e.lineno`, so I left the snippet logic as is.

Giom-V:
Thanks for the submission @rishabhsai, let me just take some time to review the new script and decide if we need it or not (I'm not sure we need more reasons to fail PRs).
@kkorpal I'm also interested in your thought (in our next sync)

Giom-V (approved):

PR #1377, opened 2026-09-15
fix(nb_tester): handle max_retries=0, avoid in-place rule mutation, and enforce per-cell timeouts

Giom-V:
## Summary
Follow-up fix to #1368 addressing code review findings and edge cases in `tools/nb_tester`:

1. **`max_retries=0` Handling & CLI Override Precedence (`rules.py`, `config.py`, `cli.py`)**:
   - Fixed falsy trap where `max_retries: 0` fell back to the default `cell_max_retries` (3).
   - Added `TesterConfig.OVERRIDE_MAX_RETRIES` so passing `--max-retries 0` (or any integer) via CLI properly takes precedence over `default_rules.yaml`.
2. **Immutable Rule Resolution (`rules.py`)**:
   - `resolve_cell_action_and_strategy()` now returns a fresh resolved `CellRule` instance rather than mutating shared rule objects in `nb_rules.cell_rules`.
3. **Per-Cell Timeout Enforcement (`executor.py`)**:
   - Explicitly updates `client.timeout = cell_timeout` before executing each code cell so per-cell `timeout_sec` overrides are respected by `NotebookClient`.
4. **Style Compliance**:
   - Wrapped all lines exceeding the 100-character limit across `cli.py`, `config.py`, `executor.py`, and `rules.py`.


tools/nb_tester/config.py
```diff
@@ -44,12 +44,22 @@ class TesterConfig:
     # User rule: If specified, trust model names and centralize all LLM model references.
     SECURITY_AUDITOR_MODEL: str = "gemini-3.1-pro-preview"
     SECURITY_AUDITOR_FALLBACKS: List[str] = field(
-        default_factory=lambda: ["gemini-3.1-pro-preview", "gemini-3.7-flash", "gemini-3.6-flash", "gemini-2.5-pro"]
+        default_factory=lambda: [
+            "gemini-3.1-pro-preview",
+            "gemini-3.7-flash",
```
Giom-V, line 50: Can you add 3.8 in the list

tools/nb_tester/config.py
```diff
@@ -44,12 +44,22 @@ class TesterConfig:
...
+            "gemini-2.5-pro",
+        ]
     )
 
     OUTPUT_JUDGE_MODEL: str = "gemini-3.5-flash-lite"
     OUTPUT_JUDGE_FALLBACKS: List[str] = field(
-        default_factory=lambda: ["gemini-3.5-flash-lite", "gemini-3.7-flash", "gemini-3.6-flash", "gemini-2.5-flash"]
+        default_factory=lambda: [
+            "gemini-3.5-flash-lite",
+            "gemini-3.7-flash",
```
Giom-V, line 61: add 3.8

tools/nb_tester/config.py
Giom-V: add 3.8 as well

